PR #1, opened 2020-08-11
Copyedit readme following Fivetran docs style

fivetran-chloe:
Updated language to match other READMEs.

I've also left a few questions inline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Check [dbt Hub](https://hub.getdbt.com/) for the latest installation instruction
 
 ## Configuration
 Because the Fivetran Log connector exists at the *destination* level, you will need to declare each destination's log connector as a separate source in `src_fivetran_log.yml`. 
-
+ 
 However, because each schema is identical in table structure, you can use [yaml anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) to avoid duplicating code. You'll need to provide the structure in the first source, which you can then point to in subsequent ones. See the example configuration of two sources below:
 
 ```yml
--- a/README.md
+++ b/README.md
@@ -1,18 +1,18 @@
 # Fivetran Log 
 
-This package models Fivetran Log data from [our free, internal connector](https://fivetran.com/docs/logs/fivetran-log). It uses **destination-level** data in the format described by [this ERD](https://docs.google.com/presentation/d/1lny-kFwJIvOCbKky3PEvEQas4oaHVVTahj3OTRONpu8/?usp=sharing) and unions the data to the **account level**.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,7 +21,7 @@ Thus, the package's main foci are to:
 | fivetran\_log\_connenector\_status        | Each record represents a connector loading data into a destination, enriched with data about the connector's status and the status of its data flow.                                          |
 | fivetran\_log\_transformation\_status     | Each record represents a transformation, enriched with data about the transformation's last sync and any tables whose new data triggers the transformation to run. |
 | fivetran\_log\_mar\_table\_history     | Each record represents a table's active volume for a month, complete with data about its connector and destination.                             |
-| fivetran\_log\_credit\_mar\_history    | Each record represents a destination's consumption, via its MAR, total credits used, and credits per millions MAR.                             |
+| fivetran\_log\_credit\_mar\_history    | Each record represents a destination's consumption by showing its MAR, total credits used, and credits per millions MAR.                             |
```
fivetran-chloe, line 24: Just a note: We try to avoid using "via" in our docs because it has several different definitions ("by way of/through" or "by means of")

README.md
```diff
@@ -70,10 +70,10 @@ sources:
 ```
 
 Then, in each of the staging models, the `union_source_tables(table_name)` macro will:
-1. iterate through the declared sources and their tables
-2. check if the source's database does indeed have a relation matching the given `table_name` (necessary because the `transformation` and `trigger_table` tables will only exist if you've created a transformation in that destination)
-3. union the matching tables
-4. in the unioned table, store the record's source's *database* as `destination_database`
+1. Iterate through the declared sources and their tables
+2. Verify that the source's database has a relation matching the given `table_name`. The relation is necessary because the `transformation` and `trigger_table` tables will only exist if you've created a transformation in that destination.
```
fivetran-chloe, line 74: Is "the relation" the correct subject of the second sentence?

README.md
```diff
@@ -21,15 +21,15 @@ Thus, the package's main foci are to:
...
 
 
 ## Installation Instructions
 Check [dbt Hub](https://hub.getdbt.com/) for the latest installation instructions, or [read the dbt docs](https://docs.getdbt.com/docs/package-management) for more information on installing packages.
 
 ## Configuration
 Because the Fivetran Log connector exists at the *destination* level, you will need to declare each destination's log connector as a separate source in `src_fivetran_log.yml`. 
-
+ 
 However, because each schema is identical in table structure, you can use [yaml anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) to avoid duplicating code. You'll need to provide the structure in the first source, which you can then point to in subsequent ones. See the example configuration of two sources below:
```
fivetran-chloe, line 33: What do you mean by "subsequent ones"? Subsequent sources? Subsequent structures?

README.md
fivetran-jamie: Hm yeah that doesn't make sense. It should probably be "this verification step" or something along those lines

README.md
fivetran-jamie: Sources -- I'll clarify that!

README.md
fivetran-jamie: Ah that makes sense -- good to know!

PR #10, opened 2021-03-10
refactoring

fivetran-jamie:
- add double underscores to model names
- remove multi-warehouse logic and `destination_database` columns
- add custom schemas
- use enablement variables for transformations and trigger_tables 
- incorporate `log.connector_id` improvement (used to erroneously use the connector name) 
- add audit_table and schema_changelog models
> note: the `active_volume` table still has the connector_id/name issue


models/fivetran_log__connector_status.sql
```diff
@@ -56,9 +54,9 @@ connector_metrics as (
         max(case when event_type = 'WARNING' then connector_log.created_at else null end) as last_warning_at
 
     from connector_log 
-        right join connector on connector_log.connector_name = connector.connector_name
-        and connector_log.destination_database = connector.destination_database
-    group by 1,2,3,4,5,6,7
+        right join connector
```
fivetran-kristin, line 57: why right join instead of left?

```
from connector
left join connector_log
```

fivetran-joemarkiewicz (changes requested):
Looks great, I just had a few comments! Some of my comments are questions so we can discuss during hackathon hours if you would like.

models/fivetran_log__connector_status.sql
fivetran-jamie: i think truly no particular reason.... should i change to left for consistency?

Issue #14, opened 2021-04-04
Package is failing on Snowflake, connector_type should be connector_type_id


Martin-Carlsson:
**Package is failing on Snowflake**

In [stg_fivetran_log__connector.sql](https://github.com/fivetran/dbt_fivetran_log/blob/master/models/staging/stg_fivetran_log__connector.sql), the column name `connector_type` should be `connector_type_id`.

**Schema on Snowflake for fivetran_log:**
![image](https://user-images.githubusercontent.com/7769335/113519672-c35dae00-959e-11eb-88d8-556a154a43c3.png)

fivetran-joemarkiewicz:
Hi @Martin-Carlsson thanks for opening this issue. I am sorry to see that your dbt job is failing due to this bug. It seems this is a result of the fivetran_log connector updating the `connector_type` field to now be `connector_type_id`.

I will investigate a solution for this and respond back here with my proposed solution and then with the details around the timing of a new release to be cut for the package with the bug fix included.

Thank you!

fivetran-joemarkiewicz:
Hi @Martin-Carlsson thanks for opening this issue. I am sorry to see that your dbt job is failing due to this bug. It seems this is a result of the fivetran_log connector updating the CONNECTOR table schema to no longer provide the `connector_type` field and now only include `connector_type_id` field.

I will investigate a solution for this and respond back here with my proposed solution and then with the details around the timing of a new release to be cut for the package with the bug fix included.

Thank you!

fivetran-joemarkiewicz:
Hi @Martin-Carlsson I was able to leverage macros that we use within our other packages to fix the bug you are seeing. I have a the above PR opened which incorporates this fix. I will keep you updated on the progress.

fivetran-joemarkiewicz:
Hi @Martin-Carlsson thank you again for opening this issue and bringing our attention to the bug. We have since released the new v0.3.1 version of the [fivetran_log](https://hub.getdbt.com/fivetran/fivetran_log/latest/) package. You may update your dependency within your `packages.yml` file to the below:

```yml
packages:
  - package: fivetran/fivetran_log
    version: 0.3.1
```

PR #16, opened 2021-04-05
Connector type id fix

fivetran-joemarkiewicz:
This PR includes the following updates:

- Addresses Issue #14 by adding the `get_connector_columns` macro and `stg_fivetran_log__connector_tmp` model to the package in order to reflect the connector level change that resulted in removing the `connector_type` field and now only contains `connector_type_id`.
- Package version upgrade to v0.3.1

The addition of the macro allows for customers that are still using a previous version of the fivetran_log connector to use the package as well as allowing new fivetran_log connector users to successfully use the package without run failures due to the `connector_type` field not being included.

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -1,7 +1,7 @@
 config-version: 2
 
 name: 'fivetran_log'
-version: '0.3.0'
+version: '0.3.1'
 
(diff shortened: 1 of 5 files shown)


fivetran-jamie (approved):
looks good to me!

Issue #27, opened 2021-12-13
BUG - fivetran_log__audit_table should be partitioned


CraigWilson-ZOE:
**Are you a current Fivetran customer?** 
Craig Wilson, ZOE, Data Engineer

**Describe the bug**
We are monitoring how long each model in our dbt pipeline takes to process and the fivetran_log__audit_table model is one of the longest running that we have. The average execution time is 490 seconds.
Looking at the code I believe this model would benefit from being partitioned and processing only the latest day, rather than all data.

**Steps to reproduce**
1. Run the dbt fivetran_log package.
2. In the screenshot you can see the average run time per day for the package, and you can see this is steadily increasing slightly. Looking at this trend I would think it would increase indefinitely.
<img width="1072" alt="Screenshot 2021-12-13 at 13 35 20" src="https://user-images.githubusercontent.com/80682772/145822023-506bf11c-9d0d-4caa-9829-6818149777df.png">

**Expected behavior**
I would expect the time of execution to be more constant and not be as high as it currently is

**Project variables configuration**
only copying configuration for the relevant section due to security.
```yml
    # Fivetran log package configuration
    fivetran_log:
      fivetran_log_database: xxxxxx    # hidden for security
      fivetran_log_schema: fivetran_log
      fivetran_log_using_transformations: false # this will disable all transformation + trigger_table logic
      fivetran_log_using_triggers: false # this will disable only trigger_table logic
```

**Package Version**

```yml
packages:
  # includes dbt_utils, thus no need to seperately import it
  - package: calogica/dbt_date
    version: [">=0.4.0", "<0.5.0"]
  - package: fivetran/mixpanel
    version: [">=0.4.0", "<0.5.0"]
  - package: calogica/dbt_expectations
    version: [">=0.4.0", "<0.5.0"]
  - package: dbt-labs/codegen
    version: 0.4.0
  - package: data-mie/dbt_profiler
    version: 0.1.4
  - package: fivetran/stripe_source
    version: 0.4.3
  - package: fivetran/stripe
    version: 0.5.0
  - package: fivetran/fivetran_log
    version: [">=0.4.0", "<0.5.0"]
```

**Warehouse**
- [x] BigQuery
- [ ] Redshift
- [ ] Snowflake
- [ ] Postgres
- [ ] Databricks
- [ ] Other (provide details below)

**Additional context**
N/A

**Screenshots**
Attached higher up

**Please indicate the level of urgency** 
[...]

PR #29, opened 2021-12-21
Initial incremental setup for BigQuery

DylanBaker:

Partitions the audit log table and makes it incremental (currently just on BigQuery).

- [X] No  (please provide explanation how the change is non breaking below.)

**Is this PR in response to a previously created Issue**

- [X] Yes, #27 
- [ ] No 

- [ ] CircleCi  
- [X] Other (please provide additional testing details below)

Locally on BigQuery

- [X] BigQuery
- [ ] Redshift
- [ ] Snowflake
- [ ] Postgres
- [ ] Databricks

### DIFF
--- a/models/fivetran_log__audit_table.sql
+++ b/models/fivetran_log__audit_table.sql
@@ -1,3 +1,19 @@
+{{ config(
+    materialized='incremental',
+    partition_by={
+        'field': 'sync_start',
+        'data_type': 'timestamp',
+        'granularity': 'day'
+    }
+) }}
+
+{%- call statement('max_sync_start', fetch_result=True) -%}
+    select date(max(sync_start)) from {{ this }}
+{%- endcall -%}
+
+{%- set query = load_result('max_sync_start') -%}
+{%- set max_sync_start = query['data'][0][0] -%}
+
 with sync_log as (
     
     select 
@@ -7,6 +23,10 @@ with sync_log as (
     from {{ ref('stg_fivetran_log__log') }}
 
     where event_subtype in ('sync_start', 'sync_end', 'write_to_table_start', 'write_to_table_end', 'records_modified')
+
+    {% if is_incremental() %}
+    and date(created_at) >= '{{ max_sync_start }}'
+    {% endif %}
 ),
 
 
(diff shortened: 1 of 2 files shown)


models/fivetran_log__audit_table.sql
```diff
@@ -1,3 +1,19 @@
...
+    }
+) }}
+
+{%- call statement('max_sync_start', fetch_result=True) -%}
+    select date(max(sync_start)) from {{ this }}
+{%- endcall -%}
+
+{%- set query = load_result('max_sync_start') -%}
+{%- set max_sync_start = query['data'][0][0] -%}
+
```
fivetran-jamie, line 16: @DylanBaker i've never seen this way to limit to the new increments -- is this (having a call statement, referring to that query output in the `is_incremental()` block) more performant or anything compared to having a subquery straight in the `is_incremental` block? or more of a stylistic choice?

models/staging/stg_fivetran_log__log.sql
```diff
@@ -1,3 +1,11 @@
+{{ config(
+    partition_by={
```
fivetran-jamie, line 2: by default this model is a view -- would partitions make a difference here?

models/fivetran_log__audit_table.sql
DylanBaker: On everything that isn't BigQuery, it's a stylistic choice (and possibly even a less performant one, but on BigQuery a subquery doesn't cause a partition to get pruned correctly, so you don't benefit at all from the incremental logic from a query cost point of view. (i.e. it still scans the whole table even if the subquery would tell you to do otherwise.) It needs a literal value. It's very dumb.

models/staging/stg_fivetran_log__log.sql
DylanBaker: No, it wouldn't. I actually just did this here to test, because I couldn't actually partition the underlying raw table, which is what we could have people do in reality. This will/should get removed before we merge.

models/fivetran_log__audit_table.sql
fivetran-jamie: oh wow i didn't know that! so we should probably be employing this kind of incremental logic across the board the board ...

PR #31, opened 2021-12-29
Audit table incrementality  + Postgres compatibility

fivetran-jamie:

fivetran made PR

a couple of things:
1. draws from Dylan's PR (https://github.com/fivetran/dbt_fivetran_log/pull/29) to make the audit table incremental (and partition the staging `log` table)
2. merged Brice Luu's PR (https://github.com/fivetran/dbt_fivetran_log/pull/30) to explicitly cast the start date of the date spine in `connector_daily_events` to avoid string-type errors on postgres
3. adds postgres to circle ci testing
3. a minor tweak to use dbt_utils.group_by() 

- [x] Yes (please provide breaking change details below.)
- [ ] No  (please provide explanation how the change is non breaking below.)

mmm unsure if this is a breaking change or not. a pre-existing table now has a different materialization (incremental) but i believe users do _not_ need to explicitly run a full refresh to capture this new change.

**Is this PR in response to a previously created Issue**

- [x] Yes, Issue https://github.com/fivetran/dbt_fivetran_log/issues/27 (also this partly addresses the postgres compatibility part of https://github.com/fivetran/dbt_fivetran_log/issues/25) .... wondering if we should include postgres circle ci testing here....
- [ ] No 

- [x] CircleCi  
- [x] Other (please provide additional testing details below) -> tested locally on BQ

- [x] BigQuery
- [x] Redshift
- [x] Snowflake
- [x] Postgres
- [ ] Databricks

🥱 

**Feedback**


fivetran-joemarkiewicz:
The beauty of the dbt_fivetran_log latest release being a pre-release is that the next release will be breaking for v1 compatibility. Therefore, we can include this in the next official breaking release!

models/fivetran_log__audit_table.sql
```diff
@@ -106,7 +122,7 @@ sum_records_modified as (
         and records_modified_log.created_at > limit_to_table_starts.sync_start 
         and records_modified_log.created_at < coalesce(limit_to_table_starts.sync_end, limit_to_table_starts.next_sync_start) 
 
-    group by 1,2,3,4,5,6,7,8,9
+    {{ dbt_utils.group_by(n=9) }}
```
fivetran-joemarkiewicz, line 125: ❤️

models/fivetran_log__audit_table.sql
```diff
@@ -1,3 +1,14 @@
+{{ config(
+    materialized='incremental',
+    partition_by={
+        'field': 'sync_start',
+        'data_type': 'timestamp',
+        'granularity': 'day'
```
fivetran-joemarkiewicz, line 6: Is day the ideal granularity to partition by? I remember with Zendesk and Jira we encountered some instances where we were over partitioning. Can we confirm that this is the best partition to use?

models/staging/stg_fivetran_log__log.sql
```diff
@@ -1,3 +1,11 @@
+{{ config(
+    partition_by={
+        'field': 'created_at',
+        'data_type': 'timestamp',
+        'granularity': 'day'
```
fivetran-joemarkiewicz, line 5: Same comment as above. Can we ensure `day` is the ideal grain to partition by.

fivetran-joemarkiewicz (changes requested):
Thanks so much for tackling this @fivetran-jamie 🙌 

I have a few minor comments about ensuring `day` is the correct grain to partition by. It seems to be working great when I test locally. I know we encountered issues in other packages where we over-partitioned and I want to ensure this is the ideal partition to roll out.

Additionally, a few other updates that are not inline which we should apply before merging:
- [ ] Update the `integration_tests/dbt_project.yml` to have the same versioning as the package
- [ ] Update the CHANGELOG.md for the `v0.5.0` release notes.

models/fivetran_log__audit_table.sql
fivetran-joemarkiewicz: Can we ensure we do not need a merge key in this incremental strategy? 🤔 

I worry that not using a merge key could result in duplicate rows.

fivetran-joemarkiewicz: Maybe we consider insert-override?

fivetran-jamie: added a merge key! and looking at log connector data, it looks like it only includes data from after it was set up (which earliest it could be for anyone is [march 2020](https://fivetran.com/docs/logs/fivetran-log/changelog#march2020) when the active_volume table was created. so i think no one would reach the 4k+ daily partition limit

models/fivetran_log__audit_table.sql
```diff
@@ -7,13 +19,18 @@ with sync_log as (
     from {{ ref('stg_fivetran_log__log') }}
 
     where event_subtype in ('sync_start', 'sync_end', 'write_to_table_start', 'write_to_table_end', 'records_modified')
+
+    {% if is_incremental() %}
+        and created_at >= ( select max(sync_start) from {{ this }} )
+    {% endif %}
```
fivetran-joemarkiewicz, line 25: Based off @DylanBaker comment [here](https://github.com/fivetran/dbt_fivetran_log/pull/29#discussion_r778713946) I would assume we should be taking a different approach here. Otherwise thee incremental strategy essentially does nothing, if I am understanding this correctly 🤷 

Based off Dylan's comment, it looks like BigQuery will do a full table scan with this logic. That will defeat the purpose and I feel we should take the other approach so to address the query cost issue.

models/staging/stg_fivetran_log__log.sql
```diff
@@ -1,3 +1,11 @@
...
+    }
+) }}
+
```
fivetran-joemarkiewicz, line 8: We should remove this as it was just for testing purposes for [Dylan](https://github.com/fivetran/dbt_fivetran_log/pull/29#discussion_r778714632).

Instead we should maybe add documentation to the CHANGELOG to highlight how users can add a partition on an existing table.

fivetran-joemarkiewicz (changes requested):
A few additional questions based off Dylan's response to your questions.

models/fivetran_log__audit_table.sql
fivetran-jamie: swapped this out! wondering if we should switch the logic in other packages with incremental models?

models/staging/stg_fivetran_log__log.sql
fivetran-jamie: removed the code! would pointing folks to the fivetran [docs](https://fivetran.com/docs/destinations/bigquery/partition-table) on how to convert to partitioned tables fit what you're thinking of?

models/fivetran_log__audit_table.sql
fivetran-jamie: also hitting some errors on circle ci but not locally?? `max_sync_start` is `None` on incremental runs and that's failing :/

models/staging/stg_fivetran_log__log.sql
fivetran-joemarkiewicz: Yeah that would be perfect!

models/fivetran_log__audit_table.sql
fivetran-joemarkiewicz: Ahhh I think this is because the table it is trying to reference doesn't exist yet. We may need to leverage the --depends-on comment for this?

fivetran-joemarkiewicz: Essentially my understanding is this is being executed during the first compile, and since CircleCi is starting with a fresh environment, the `sync_start` doesn't technically exist yet. We need to explicitly tell dbt where it can find the sync_start during compile time.

fivetran-joemarkiewicz: I am thinking we may need to do something like this:
https://github.com/fivetran/dbt_zendesk/blob/5922ae9df394fde85e8c3fa533ac2707bc5fa78a/models/utils/int_zendesk__calendar_spine.sql#L1-L23

fivetran-jamie: ok got this working with some various if statements...

models/fivetran_log__audit_table.sql
```diff
@@ -106,8 +134,18 @@ sum_records_modified as (
...
-    group by 1,2,3,4,5,6,7,8,9
+    {{ dbt_utils.group_by(n=9) }}
+
+),
+
+surrogate_key as (
+
+    select 
+        *,
+        {{ dbt_utils.surrogate_key(['connector_id', 'destination_id', 'table_name', 'write_to_table_start']) }} as unique_table_sync_key
```
fivetran-joemarkiewicz, line 145: Can you add the new `unique_table_sync_key` to the fivetran_log.yml for this table and regenerate the docs.

fivetran-joemarkiewicz: Also, maybe we can remove the unique combination of columns test and instead simplify the test logic by just applying a `unique` and `not_null` test on the new field.

fivetran-joemarkiewicz (changes requested):
Thanks for applying the changes @fivetran-jamie! I tested this locally by limiting the days in the staging model, running a `--full-refresh` and then removing my limit and running a normal `dbt run`. I found the merge worked as intended and didn't see any duplicate records. I think the merge key was the "key" here (pun intended). Thanks!

I have just a few quick comments in line and highlighted below. Once those changes are made then I will approve this PR!

- [ ] Be sure to add Postgres compatibility to the README
- [ ] Can you add some inline comments to the incremental logic within the `fivetran_log__audit_table` model. I imagine customers (and us in the future) will have a hard time understanding what exactly is happening here. Some quick comments will make that easier!
- [ ] Add the `unique_table_sync_key` to the `fivetran_log.yml` and regenerate the docs.
- [ ] Consider defining the test for the unique key instead of the combination for simplicity.

fivetran-joemarkiewicz (approved):
Looks great! Thanks so much @fivetran-jamie 😄. Let's hold off merging until @gareginordyan gives a review.

One quick question - is there a reason for the extra files in the `docs/` folder? Usually all we need is the `index.html`, `catalog.json`, and the `manifest.json`. Curious why the other files are included in this PR? Can you remove them if they are not necessary.

models/fivetran_log__audit_table.sql
```diff
@@ -7,13 +19,31 @@ with sync_log as (
...
+
+    {%- set query = load_result('max_sync_start') -%}
+
+    -- store it as a singular value
+    {%- set max_sync_start = query['data'][0][0] -%}
+
+        -- compare the new batch of data to the latest sync already stored in this model
+        and date(created_at) >= '{{ max_sync_start }}'
+
+    {% endif %}
```
fivetran-joemarkiewicz, line 38: Only comment here would be to update the names of the variables to be a bit more clear. See my suggestions below:
```suggestion
    {%- call statement('max_sync_start', fetch_result=True) -%}
        select date(max(sync_start)) from {{ this }}
    {%- endcall -%}

    -- load the result from the above query into a new variable
    {%- set query_result = load_result('max_sync_start') -%}

    -- the query_result is stored as a dataframe. Therefore, we want to now store it as a singular value.
    {%- set max_sync_start = query_result['data'][0][0] -%}

        -- compare the new batch of data to the latest sync already stored in this model
        and date(created_at) >= '{{ max_sync_start }}'

    {% endif %}
```

CHANGELOG.md
```diff
@@ -1,3 +1,17 @@
...
+  - For help upgrading your dbt project to dbt v1.0.0, I recommend reviewing dbt-labs [upgrading to 1.0.0 docs](https://docs.getdbt.com/docs/guides/migration-guide/upgrading-to-1-0-0) for more details on what changes must be made.
+- Upgrades the package dependency to refer to the latest `dbt_fivetran_utils`. The latest `dbt_fivetran_utils` package also has a dependency on `dbt_utils` [">=0.8.0", "<0.9.0"].
+  - Please note, if you are installing a version of `dbt_utils` in your `packages.yml` that is not in the range above then you will encounter a package dependency error.
+
+## Additional Features
+- Materializes the `fivetran_log__audit_table` incrementally, and employs partitioning for BigQuery users. As a non-incremental table, this model involved high runtimes for some users ([#27](https://github.com/fivetran/dbt_fivetran_log/issues/27))
+  - If you would like to apply partitioning to the underlying source tables (ie the `LOG` table), refer to [Fivetran docs](https://fivetran.com/docs/destinations/bigquery/partition-table) on how to do so.
```
fivetran-joemarkiewicz, line 12: Can you actually also include a version of this in the README. This way we can make it more apparent to first time users as well that they can partition the base table being synced by Fivetran to improve performance and cost even more.

fivetran-joemarkiewicz (approved):
@fivetran-jamie thanks again for all your hard work on this update!

I chatted with @gareginordyan and as the `insert_overwrite` logic is not compatibile with all our supported destinations, it makes sense to keep the merge incremental strategy. I did make a few inline comments that we should apply before we merge this PR and cut a new release.
- [x] Add details to the README on how the customer can add the partition to their base table in order to improve performance and cost even more. As it stands currently, the _time_ of the model is greatly reduced with our updates. However, if they add the partition logic like you highlighted in the CHANGELOG then they will see a better impact on the _cost_ of the query.
- [ ] Rename a few of the variables in the incremental logic to read a bit easier. I added a suggestion.

Once you make these changes, then we will be good to merge this PR and cut the release!

CHANGELOG.md
fivetran-jamie: added!

Issue #39, opened 2022-02-25
[Bug] Broken connector reported as "connected"


carlioth:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

In the Fivetran UI we can see that we have a connector (Salesforce sandbox) that is broken: 
![Fivetran-ui-broken](https://user-images.githubusercontent.com/908670/155677968-9e79b259-a8d0-4006-8372-9b31a23d5214.png)

We also have our own System status dashboard where we rely on the data coming out of the models in this dbt package. 
The issue here is that it's showing that this connector is healthy, described as "connected"

![row-model-connected](https://user-images.githubusercontent.com/908670/155678255-84b77268-034e-4f88-9dfe-0b80ae8c021a.png)

I quickly started to dig into what might be going wrong, and perhaps here is something. If I run: 

```sql
    select *
    from "DB"."SCHEMA"."stg_fivetran_log__log"

    where event_type = 'SEVERE'
        or event_subtype like 'sync%'
        and connector_id = 'crusade_plausible'
        order by CREATED_AT desc
```

I'm getting a sync_start then a Severe log, and then a sync_end log. 

![rows-from-stg-fivetran-log-log](https://user-images.githubusercontent.com/908670/155679592-1084b8a0-ec0b-4bfd-a8ef-e2544808b01c.png)

When this later is categorised, this might be where the issue is presented? 

https://github.com/fivetran/dbt_fivetran_log/blob/691b82a06eb22a940b628dbac428b2f16b1ae9d8/models/fivetran_log__connector_status.sql#L83-L101

For us this is a high priority issue since we rely on the data from our System status dashboard where we monitor all our Fivetran connectors. 

### Expected behavior

We expect the model `fivetran_log__connector_status` to report the connector as broken if it's broken in the Fivetran UI. 

### dbt Project configurations

models:
  +persist_docs:
    relation: true
    columns: true
  data_eng:
    sources:
      +materialized: table
  fivetran_log:
    +schema: # leave these blank to use the target_schema
    staging:
      +schema: # leave these blank to use the target_schema

vars:
  fivetran_log:
    fivetran_log_database: db
    fivetran_log_schema: schema

### Package versions

packages:
  - package: fivetran/fivetran_log
    version: 0.5.0

### What database are you using dbt with?

snowflake

### dbt Version

Dbt version 1.0.0

### Are you willing to open a PR to help address this issue?

- [ ] Yes.
[...]

PR #41, opened 2022-03-17
Bugfix/connector status

fivetran-joemarkiewicz:

Fivetran created PR

- Per the [Fivetran Log December 2021 Release Notes](https://fivetran.com/docs/logs/changelog#december2021) every sync results in a final `sync_end` event. In the previous version of this package, a successful sync was identified via a `sync_end` event while anything else was a version of broken. Since all syncs result in a `sync_end` event now, the package has been updated to account for this change within the connector.
- To account for the above fix, a new field (`last_successful_sync_completed_at`) was added to the `fivetran_log__connector_status` model. This field captures the last successful sync for the connector.

- [X] Yes

- [X] No  (please provide an explanation as to how the change is non-breaking below.)

This will simply adjust the logic for an existing field to capture successful and broken connectors as well as add a new field. This will not be breaking for existing users.

- [X] Yes

- [X] Yes, Issue/Feature #39  
- [ ] No 

- [X] CircleCi  

Tested this locally and received confirmation from the customer that this resolved their issue.

- [X] BigQuery
- [X] Redshift
- [X] Snowflake
- [X] Postgres
- [X] Databricks

🧾 

**Feedback**


models/fivetran_log__connector_status.sql
```diff
@@ -1,14 +1,19 @@
...
     from {{ ref('stg_fivetran_log__log') }}
 
     -- only looking at errors, warnings, and syncs here
     where event_type = 'SEVERE'
         or event_type = 'WARNING'
         or event_subtype like 'sync%'
+        or (event_subtype = 'status' 
+            and {{ fivetran_utils.json_parse(string="message_data", string_path=["status"]) }} ='SUCCESSFUL')
         or (event_subtype = 'status' 
             and {{ fivetran_utils.json_parse(string="message_data", string_path=["status"]) }} ='RESCHEDULED'
```
fivetran-jamie, line 15: wondering if we should combine these two to only call `json_parse` once 

so like 
```suggestion
        or (event_subtype = 'status' 
            and {{ fivetran_utils.json_parse(string="message_data", string_path=["status"]) }}  in ('SUCCESSFUL', 'RESCHEDULED'))
```

models/fivetran_log__connector_status.sql
```diff
@@ -58,13 +63,23 @@ connector_metrics as (
         max(case when connector_log.event_subtype = 'sync_end' 
             then connector_log.created_at else null end) as last_sync_completed_at,
 
+        max(case when connector_log.event_subtype = 'status'
+                and {{ fivetran_utils.json_parse(string="connector_log.message_data", string_path=["status"]) }} ='SUCCESSFUL'
+            then connector_log.created_at else null end) as last_successful_sync_completed_at,
```
fivetran-jamie, line 68: new customers will have the important json status stuff in `sync_end` events, and it sounds like we'll be retroactively rolling that out to old customers in the future as well. i think the following could capture what we need for everyone? 
```suggestion
        max(case when connector_log.event_subtype in ('status', 'sync_end')
                and {{ fivetran_utils.json_parse(string="connector_log.message_data", string_path=["status"]) }} ='SUCCESSFUL'
            then connector_log.created_at else null end) as last_successful_sync_completed_at,
```

models/fivetran_log__connector_status.sql
```diff
@@ -95,7 +110,8 @@ connector_health as (
             when last_sync_completed_at is null and last_error_at is null then 'initial sync in progress'
 
             -- there's been an error since the connector last completed a sync
-            when last_error_at > last_sync_completed_at then 'broken'
+            when --last_error_at > last_sync_completed_at and 
+                last_sync_batch_id = last_error_batch then 'broken'
```
fivetran-jamie, line 114: reminder to remove commented out code

i'd also maybe change the comment in line 112 to like `the last attempted sync had an error`

models/fivetran_log.yml
```diff
@@ -33,14 +33,18 @@ models:
...
+
       - name: last_sync_started_at
         description: > 
           Timestamp of when the last sync was started. Note that if new data comes in after a sync has 
           begun, the new data will not be included in this sync.
 
       - name: last_sync_completed_at
         description: > 
-          Timestamp of when the last sync was complted.
+          Timestamp of when the last sync was completed. This could be either a failure or a successful sync.
```
fivetran-jamie, line 47: what about `Timestamp of when the last sync attempt was completed...`

models/fivetran_log__connector_status.sql
```diff
@@ -114,6 +130,7 @@ connector_recent_logs as (
         connector_health.connector_type,
         connector_health.destination_id,
         connector_health.connector_health,
+        connector_health.last_successful_sync_completed_at,
         connector_health.last_sync_started_at,
         connector_health.last_sync_completed_at,
```
fivetran-jamie, line 135: is `last_sync_completed_at` valuable to have anymore now that there is `last_successful_sync_completed_at`? 

or like should we change the name to `last_sync_attempt_completed_at` to better differentiate those

models/fivetran_log.yml
fivetran-joemarkiewicz: Yeah I think I over explained this one haha I think before we alluded to the fact that the timestamp being completed meant it succeeded, but that is not the case anymore. I like the idea of keeping it simple.

models/fivetran_log__connector_status.sql
fivetran-joemarkiewicz: Way more elegant with this method. Thanks!

models/fivetran_log__connector_status.sql
fivetran-joemarkiewicz: I was struggling with this one. I wanted to remove it, but in terms of the connector itself a sync can complete and not succeed. So I could see it being valuable to see the last successful sync vs the last completed sync to see if your connector has just been failing for the last few days.

models/fivetran_log__connector_status.sql
fivetran-joemarkiewicz: Actually I just tried this and found that it was not working as intended. The `SUCCESSFUL` status events do not also have the `intended behavior` message attached to them. Because of that, when we combine these json parse queries, they actually do not bring through the successful events. 

Because of that, we should keep it as is.

models/fivetran_log__connector_status.sql
fivetran-joemarkiewicz: I feel we should keep it, but that is why I made sure to add the verbiage to the yml to highlight that it doesn't indicate a successful sync. What are your thoughts?

fivetran-joemarkiewicz (commented):
@fivetran-jamie I went through and addressed your comments and review notes to my changes. Thanks so much for giving this review 🙏 

I did have a few points I decided to keep my original logic and have questions I want your input on. Let me know what your thoughts are!

models/fivetran_log__connector_status.sql
fivetran-jamie: actually yah i agree....like a sync can be partially successful right? could be worth it to know if some of your tables got written to while others did not

models/fivetran_log.yml
fivetran-jamie: oh no i think we should keep the second sentence! the `...` included that haha

fivetran-jamie: so like `Timestamp of when the last sync attempt was completed. This could be either a failure or a successful sync.`

models/fivetran_log__connector_status.sql
fivetran-jamie: and yeah i think your original yml description may highlight the differences enough

models/fivetran_log.yml
fivetran-joemarkiewicz: Ahhh the word attempt is what I was missing here. Yeah I like the original haha thanks!

fivetran-jamie: oh and a reminder to regenerate the docs afterward!

models/fivetran_log__connector_status.sql
```diff
@@ -58,13 +63,23 @@ connector_metrics as (
...
-        max(case when connector_log.event_subtype = 'status'
+        max(case when connector_log.event_subtype in ('status', 'sync_end')
+                and {{ fivetran_utils.json_parse(string="connector_log.message_data", string_path=["status"]) }} ='SUCCESSFUL'
+            then connector_log.created_at else null end) as last_successful_sync_completed_at,
+
+
+        max(case when connector_log.event_subtype = 'sync_end' 
+            then connector_log.sync_batch_id else null end) as last_sync_batch_id,
+
+        max(case when connector_log.event_subtype in ('status', 'sync_end')
```
fivetran-jamie, line 74: ah good catch here too

fivetran-jamie (approved):
looks great to me! good to merge after the yml tweak

models/fivetran_log.yml
```diff
@@ -33,14 +33,18 @@ models:
       - name: connector_health
         description: Status of the connector's data flow. Can be `"broken"`, `"incomplete"`, `"connected"`, `"paused"`, `"initial sync in progress"`, or `"priority first sync"`.
 
+      - name: last_successful_sync_completed_at
+        description: > 
+          Timestamp of when the last successful sync completed.
```
fivetran-joemarkiewicz, line 38:
```suggestion
          Timestamp of when the last sync attempt was completed. This could be either a failure or a successful sync.
```

PR #44, opened 2022-04-01
addition of the usage cost source and logic to use the proper table

fivetran-joemarkiewicz:

Fivetran created PR

This PR does a couple things that important to call out:
- This package adds the new `usage_cost` source to the package. This source table will take the place of the `credits_used` model.
- However, the `usage_cost` model is not yet rolled out to all Fivetran Log users as this only is relevant for individuals on the new pricing model.
- Therefore, this PR incorporates some pretty unique logic to check and see if the user has the `usage_cost` source table within their schema using the `does_table_exist()` macro.
  - I am not sold on the name of the macro at the moment and will plan to move this to `fivetran_utils`.
- When running the `stg_fivetran_log__credits_used` (name will most likely change) model, the macro is called and used to update the `fivetran_log__usage_pricing` variable. The variable will be `true` if the `usage_cost` source is identified. If not, the variable will be `false`.
- If the variable is true then the model will run leveraging the `usage_cost` source. Otherwise, the `credits_used` source will be used.
- Downstream models and yml docs were updated to account for these changes. 

- [ ] Yes

Weirdly enough I do not believe this will be a breaking change... but maybe we should make it breaking as it will be a big update?

- [ ] Yes

- [ ] Yes, Issue/Feature [link bug/feature number here]
- [X] No 

- [ ] CircleCi  

** FYI Seed data and integration tests have not been added yet. They need to be added before moving out of draft **
- [ ] BigQuery
- [ ] Redshift
- [ ] Snowflake
- [ ] Postgres
- [ ] Databricks

🤯 

**Feedback**

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -28,3 +28,4 @@ vars:
     transformation: "{{ source('fivetran_log', 'transformation') }}"
     trigger_table: "{{ source('fivetran_log', 'trigger_table') }}"
     user: "{{ source('fivetran_log', 'user') }}"
+    usage_cost: "{{ source('fivetran_log', 'usage_cost') }}"
--- a/macros/does_table_exist.sql
+++ b/macros/does_table_exist.sql
@@ -1,0 +1,17 @@
+{%- macro does_table_exist(table_name) -%}
+    {%- if execute -%}
+    {%- set ns = namespace(has_table=false) -%}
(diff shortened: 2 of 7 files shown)


fivetran-joemarkiewicz:
Duplicate PR. The new one can be found within PR #50

PR #47, opened 2022-04-05
doc(fivetran log): Fix relative links in Fivetran log transform README

jlmendgom5TRAN:
Closes https://fivetran.height.app/T-228325

Replaces relative links with absolute links that do not break when rendered in our docs.


fivetran-joemarkiewicz (approved):
LGTM

PR #52, opened 2022-04-25
Adds options to which models are included

cmcau:
Pull Request
Partner
Chris McClellan
Director
Visualisedata Pty Ltd
AUSTRALIA

Adds options to dbt_project.yml to optionally load some of the models.  I wanted to use this package to show what Fivetran was doing (during the trial period), but some of the models are not available and that causes an error.  With these changes I don't get any errors and I can report on the data because the run finishes without any errors

- [x] Yes

- [x] Yes (please provide breaking change details below.)
- [ ] No  (please provide an explanation as to how the change is non-breaking below.)
I'm going to say YES to be safe.  I only thought of that now - if users don't have the vars defined then there will be an error I guess.  I haven't tested this, but I'm happy to do the work if this is an issue that needs to be fixed and this PR is going to benefit the package.

- [x] Yes

- [ ] Yes, Issue/Feature [link bug/feature number here]
- [x] No 

- [ ] CircleCi  

I tested it with all the vars properly defined in the root dbt_project.yml and it passed all tests.
I DID NOT test when the vars are not defined in the root dbt_project.yml

- [ ] BigQuery
- [ ] Redshift
- [x] Snowflake
- [ ] Postgres
- [ ] Databricks

:grinning: hopefully, but a lot of :worried: at the same time

**Feedback**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,22 @@
+# dbt_fivetran_log v0.5.5
+## Fixes
+- Added the option to disable some models that are not available depending on your Fivetran license. In dbt_project.yml you can now change these flags to disable parts of the package
+  - fivetran_log_account_membership
+  - fivetran_log_active_volume
+  - fivetran_log_credits_used
+  - fivetran_log_destination_membership
+  - fivetran_log_user
+
+  Include this in your dbt_project.yml file
+
+  ```
+  fivetran_log_account_membership: false # Disables account membership models
+  fivetran_log_active_volume: false # Disables account membership models
+  fivetran_log_credits_used: false # Disables account membership models
+  fivetran_log_destination_membership: false # Disables account membership models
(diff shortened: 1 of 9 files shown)


fivetran-joemarkiewicz:
Hey @cmcau thanks so much for opening this PR!! 

These changes make complete sense and I understand why you would want to remove these models from the run as they do not exist during the trial phase. My one question I have is regarding the value of the end models with these disabled. Are you still able to gain actionable insights in your opinion with these disabled?

cmcau:
Well, I don't know what I'm missing right ? ;) 
I have created a visualisation that shows the data in the log_audit table (so I can see Deletes, Updated and Replaced or Inserted by Connector and Table Name.  

Active Volume seems to be the most interesting one that I'm missing (but I'm only going on the name of the model)

Issue #57, opened 2022-07-22
[Bug] Source Freshness tests still execute for sources that are distabled


epapineau:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

We installed the fivetran_log package and disabled additional tables with the recommending configuration in `dbt_project.yml`.

When our next prod job executed `dbt source freshness` we received object does not exist errors for the tables that were disabled. 

### Relevant error log or model output

```shell
23:45:38  Database Error in source account_membership (models/staging/src_fivetran_log.yml)
23:45:38    002003 (42S02): SQL compilation error:
23:45:38    Object 'RAW.FIVETRAN_LOG.ACCOUNT_MEMBERSHIP' does not exist or not authorized.
23:45:38  
23:45:38  Database Error in source destination_membership (models/staging/src_fivetran_log.yml)
23:45:38    002003 (42S02): SQL compilation error:
23:45:38    Object 'RAW.FIVETRAN_LOG.DESTINATION_MEMBERSHIP' does not exist or not authorized.
23:45:38  
23:45:38  Database Error in source user (models/staging/src_fivetran_log.yml)
23:45:38    002003 (42S02): SQL compilation error:
23:45:38    Object 'RAW.FIVETRAN_LOG.USER' does not exist or not authorized.
23:45:38  Done.
```

### Expected behavior

If we've communicated we do not have the tables in our dataset, we should not try to run source freshness for them0.6.1

### dbt Project configurations

```
vars:
  fivetran_log:
    fivetran_log_database: raw
    fivetran_log_schema: fivetran_log
    fivetran_log_using_transformations: false  
    fivetran_log_using_triggers: false  
    fivetran_log_using_account_membership: false 
    fivetran_log_using_destination_membership: false 
    fivetran_log_using_user: false  
```

### Package versions

0.6.1

### What database are you using dbt with?

snowflake

### dbt Version

1.1

### Are you willing to open a PR to help address this issue?

- [ ] Yes.
- [ ] Yes, but I will need assistance and will schedule time during our [office hours](https://calendly.com/fivetran-solutions-team/fivetran-solutions-team-office-hours) for guidance
- [ ] No.

fivetran-joemarkiewicz:
Hi @epapineau! 

[...]

PR #58, opened 2022-08-03
Add config enabled for disable-able sources

epapineau:
Pull Request

Elize Papineau, Senior Analytics Engineer, dbt Labs
Yes, working on behalf of a customer using Fivetran 

Passes the configurable variables to `config: enabled` to prevent running source freshness on disabled models

- [x] Yes

It extends use of existing variables

- [x] Yes

- [x] Yes, [Issue](https://github.com/fivetran/dbt_fivetran_log/issues/57)
- [ ] No 

- [ ] CircleCi  

- [ ] BigQuery
- [ ] Redshift
- [x] Snowflake
- [ ] Postgres
- [ ] Databricks

:mask:

**Feedback**

### DIFF
--- a/models/staging/src_fivetran_log.yml
+++ b/models/staging/src_fivetran_log.yml
@@ -135,6 +135,8 @@ sources:
       - name: account_membership
         identifier: "{{ var('fivetran_log_account_membership_identifier', 'account_membership')}}"
         description: Table of the users given access to this account.
+        meta:
+          is_enabled: "{{ var('fivetran_log_using_account_membership', true) }}"
         columns:
           - name: account_id
             description: Foreign key referencing the ID of the `account` that the user belongs to.
@@ -150,6 +152,8 @@ sources:
       - name: destination_membership
         identifier: "{{ var('fivetran_log_destination_membership_identifier', 'destination_membership')}}"
         description: Table of the users given access to this destination.
+        meta:
+          is_enabled: "{{ var('fivetran_log_using_destination_membership', true) }}"
         columns:
           - name: destination_id
             description: Foreign key referencing the ID of the `destination` that the user is a member of.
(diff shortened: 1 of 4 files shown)


models/staging/src_fivetran_log.yml
```diff
@@ -135,6 +135,8 @@ sources:
       - name: account_membership
         identifier: "{{ var('fivetran_log_account_membership_identifier', 'account_membership')}}"
         description: Table of the users given access to this account.
+        meta:
```
fivetran-joemarkiewicz, line 138: @fivetran-jamie actually did some digging and found that dbt-core now has a [feature built in](https://docs.getdbt.com/reference/source-configs) to allow for the dynamic disabling of source! 

This means we don't need to use our override of the freshness macro and instead just disable the source freshness altogether! I just tested the new config and it seems to work as intended. 

Would you be able to try on your end the following update for all of the sources you modified to see if it works on your end as well?
```suggestion
        config:
```

fivetran-joemarkiewicz (changes requested):
@epapineau thanks so much for making these changes! They look great on my end, but I actually just discovered that there is a [native way](https://docs.getdbt.com/reference/source-configs) we can disable these sources without needing to use the `meta` config and override the freshness macro!

The best option would be to disable the source entirely instead of the `meta` config. Thankfully, the only change needed in your PR is to swap `meta` for `config`. Would you be able to make those changes and see if they work as intended on your end?

models/staging/src_fivetran_log.yml
epapineau: @fivetran-joemarkiewicz requested changes have been made 😎

fivetran-joemarkiewicz (approved):
@epapineau thanks so much for making the changes 😄 I made two small changes to bump the version in our integration tests project and also made a small edit to the CHANGELOG.

With that, this PR is good to go! Thanks so much for your collaboration and contribution to the package 🙌 These packages are made better by contributors like yourself adding to the code base. I will merge this and plan to cut a release later this morning.

PR #74, opened 2023-02-16
fix: aggregate by month in fivetran_log__mar_table_history

simon-stepper:

Yes, Simon Stepper, Analytics Lead at Capdesk from Carta

Fixing a bug introduced in v0.7.1 and switching the `fivetran_log__mar_table_history` back to a monthly aggregation.

- [x] Yes

Only changing the model's granularity to what it was in v0.7.0

- [x] Yes

- [x] Yes, Issue/Feature https://github.com/fivetran/dbt_fivetran_log/issues/73
- [ ] No 

- [ ] BuildKite  
Tested with our own dbt project on Snowflake

- [ ] BigQuery
- [ ] Redshift
- [x] Snowflake
- [ ] Postgres
- [ ] Databricks

⛷️ 

**Feedback**


models/fivetran_log__mar_table_history.sql
```diff
@@ -27,50 +27,37 @@ ordered_mar as (
...
-        schema_name,
-        table_name,
-        destination_id,
-        measured_month,
-        date(measured_date) as last_measured_at,
-        free_monthly_active_rows,
-        paid_monthly_active_rows,
-        (free_monthly_active_rows + paid_monthly_active_rows) as total_monthly_active_rows
-    
-    from ordered_mar
```
fivetran-joemarkiewicz, line 56: Unfortunately, we do need to keep this cte in the model. The recursive ability of Snowflake does not exist in other warehouses. Therefore, if we were to go forward with removing the cte, other warehouses would be prompted with a "free_monthly_active_rows field does not exist".

As such, I would recommend we bring this cte back and move the `total_monthly_active_rows` calculation to this cte.

models/fivetran_log__mar_table_history.sql
```diff
@@ -27,50 +27,37 @@ ordered_mar as (
...
+        sum(coalesce(case when lower(free_type) = 'paid'
             then incremental_rows
-        end, 0) as paid_monthly_active_rows,
-        coalesce(case when lower(free_type) != 'paid'
+        end, 0)) as paid_monthly_active_rows,
+        sum(coalesce(case when lower(free_type) != 'paid'
             then incremental_rows
-        end, 0) as free_monthly_active_rows
+        end, 0)) as free_monthly_active_rows,
+        (free_monthly_active_rows + paid_monthly_active_rows) as total_monthly_active_rows
```
fivetran-joemarkiewicz, line 39: Similar comment from the other. Can we move this to the `latest_mar` cte.

models/fivetran_log__mar_table_history.sql
```diff
@@ -27,50 +27,37 @@ ordered_mar as (
...
+    {{dbt_utils.group_by(5)}}
 
 ),
 
 mar_join as (
 
-    select 
-        latest_mar.*,
+    select
+        ordered_mar.*,
```
fivetran-joemarkiewicz, line 49: Since we will need to keep the latest_mar cte.
```suggestion
        latest_mar.*,
```

fivetran-joemarkiewicz (changes requested):
@simon-stepper thank you so much for your quick notice of this bug and for investigating the issue at hand and proposing a solution. 🥇 

Your changes look good on my end; however, I have a few small requests to update in order for other warehouses to be compatible before rolling out this in the next release. Let me know if you have any questions.

Additionally, when testing your changes I actually noticed that there may be occurrences within the `fivetran_log__mar_destinaton_history` model where a table can exist across multiple schemas in a destination. This still causes the test to fail. As such, I was able to adjust the test to include `schema_name` and that worked for my testing. I am curious if you encountered the same. Nevertheless, it is an addition to the next release we will want to incorporate. Here is a screenshot of the change I made locally.
image
 Let me know if you have any questions. I really appreciate your attentiveness and hope to roll these changes out as soon as possible.

simon-stepper:
@fivetran-joemarkiewicz thanks for the quick review! Would be great when you can apply these changes to make it work across warehouses, I'm Snowflake only.

We don't have a case where a table name is not unique per connector in our setup, so this didn't occur for me. But I can see that happening and would be a good addition to make the package more robust. Thanks!

fivetran-joemarkiewicz (approved):
Approving this PR with these changes in place. I will hold off merging this until we are able to get another customer to test on their end and validate the changes.

Thank you again for all your help @simon-stepper

Issue #75, opened 2023-02-16
[Bug] Runtime Error in `does_table_exist` macro


dimoschi:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

While building, compiling or generating logs in a project that uses `dbt_fivetran_log` package with a Snowflake target, a weird error pops up randomly. The error is about approximate match by `adapter.get_relation` which is used in `does_table_exist` macro/

### Relevant error log or model output

```shell
23:14:22  Encountered an error:
Runtime Error
  Compilation Error in model stg_fivetran_log__usage_cost (models/staging/stg_fivetran_log__usage_cost.sql)
    When searching for a relation, dbt found an approximate match. Instead of guessing 
    which relation to use, dbt will move on. Please delete "FINANCIAL_DB"."SHOPIFY"."ORDER", or rename it to be less ambiguous.
    Searched for: FINANCIAL_DB.SHOPIFY."ORDER"
    Found: "FINANCIAL_DB"."SHOPIFY"."ORDER"
    
    > in macro does_table_exist (macros/does_table_exist.sql)
    > called by model stg_fivetran_log__usage_cost (models/staging/stg_fivetran_log__usage_cost.sql)
```

### Expected behavior

As the searched table is totally unrelated with the usage of `does_table_exist` my expectation is that no error should exist.

### dbt Project configurations

Specifically for the package we use the following
```yaml
  fivetran_log:
    +tags: [monitoring, fivetran_log]
    +database: "{{ env_var('CURATED_DB', 'DEV_CURATED_DB') }}"
    +schema: MONITORING
    staging:
      +tags: staging
      +materialized: view
      +database: "{{ env_var('STAGING_DB', 'DEV_STAGING_DB') }}"
      +schema: MONITORING # leave blank for just the target_schema
```

### Package versions

```yaml
packages:
  - package: fivetran/xero
    version: [">=0.5.0", "<0.6.0"]
  - package: fivetran/hubspot
    version: [">=0.7.0", "<0.8.0"]
  - package: fivetran/shopify
    version: [">=0.7.0", "<0.8.0"]
  - package: fivetran/fivetran_log
    version: [">=0.7.0", "<0.8.0"]
  - package: calogica/dbt_expectations
    version: [">=0.8.0", "<0.9.0"]
  - package: dbt-labs/codegen
    version: [">=0.9.0", "<1.0.0"]
```

### What database are you using dbt with?

snowflake

### dbt Version

[...]

PR #77, opened 2023-03-06
bugfix/does-table-exist-order

fivetran-joemarkiewicz:

Fivetran created PR

## Bug Fixes
- The logic within the `does_table_exist` macro would run the source_relation check across **all** nodes. This opened dbt compile to erroneous failures in other (non fivetran_log) sources. This macro logic has been updated to only check the source_relation for the specific source in question.
- Adjusted the enabled variable used within the `stg_fivetran_log__credits_used` model to the more appropriate `fivetran_log__credits_pricing` name as opposed to `fivetran_log__usage_pricing`. This ensures a user may override the respective model enablement in isolation of each other.

## Documentation Updates
- Added documentation to support the `fivetran_log__usage_pricing` and `fivetran_log__credits_pricing` variable behaviors within the package.

- [X] Yes

- [X] No  (please provide an explanation as to how the change is non-breaking below.)

This should just be updating the logic within the does_table_exist macro and not impact any users in a way that would cause a breaking change.

- [X] Yes

- [X] Yes, Issue #75 
- [ ] No 

- [X] BuildKite  

- [X] BigQuery
- [X] Redshift
- [X] Snowflake
- [X] Postgres
- [X] Databricks

➿ 

**Feedback**


README.md
```diff
@@ -86,6 +86,18 @@ Some users may wish to exclude Fivetran error and warnings messages from the fin
...
+
+If for any reason you want to turn off the dynamic functionality of the package and use only one of the two sources (`usage_cost` or `credits_used`) you may use the respective below variables to control the package behavior. Please note, it is highly encouraged to ensure at least one of the variables is defined to `true`. The package is intended to work when both are `true` or when at least one is `true`.
+
+```yml
+vars:
+  fivetran_log__usage_pricing: true ## Dynamically checks the source at runtime to set as either true or false. May be overridden using this variable if desired.
+  fivetran_log__credits_pricing: true ## Dynamically checks the source at runtime to set as either true or false. May be overridden using this variable if desired.
```
fivetran-jamie, line 98: should the example code be setting these variables to `False`, since we're describing how to turn this behavior off?

fivetran-jamie: also the empty table still gets created if you set the variable to false. is that intended? if so, i feel like we should clarify that here, as it kinda sounds like these variables are supposed to totally **disable** the respective staging models/transformations

fivetran-jamie (commented):
macro looks and runs good, but got a question about the pricing plan variables!

README.md
fivetran-joemarkiewicz: Yeah I was going back and forth on this as well. We originally did not document this at all because of this functionality of it not _really_ being disabled. So I tried to add context around this.

Let me try and clarify a bit more. I also didn't want to show both of them set to `false` since that is the one form of the package that is not supported.

fivetran-joemarkiewicz: @fivetran-jamie I just attempted to update the wording. I am still not 100% on the wording. What are your thoughts on the latest version?

fivetran-jamie: i guess the behavior itself is a little confusing -- why would a customer explicitly elect to bring in null data? i get that if they no longer use credits they don't want any credit-related data in there, but ideally these variables would remove any credit columns or transformations

but if that's out of scope for rn, i think we can frame the point of these variables better. like "if your schema contains outdated tables reflecting old price plans that you would like the package to completely ignore, set these to false"

fivetran-joemarkiewicz: Yeah the tricky part here is that the credit model may still be applicable for some folks. There very well may be cases where people want both credit and usage cost. However, I really don't think there will be cases when people want to actively disable the models for the results of the output. The variables are probably more important in case the dynamic functionality doesn't function as intended.

As such, I wonder if it is better for this to be framed more as an informative section around the existence of the variables and less about how a customer would use them (since that is not really the intended outcome).

fivetran-jamie: i like it!

fivetran-joemarkiewicz:
Thanks @fivetran-jamie! If all looks good, would you mind approving the PR?

fivetran-jamie (approved):

fivetran-jamie:
approved!

Issue #93, opened 2023-10-17
[Feature] Deprecate Transformations and Trigger Staging Model and Downstream References


fivetran-joemarkiewicz:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Describe the Feature

Currently the `transformation` and `trigger` models are used for basic SQL transformations. However, these have since been deprecated in the Fivetran platform. Therefore, there is no need for these tables to be pulled through as staging models or through the transformations status end model.

Upon it's removal we will be able to rely on the `log` table to generate all status' for Fivetran dbt orchestrated transformations.

### Are you interested in contributing this feature?

- [X] Yes.
- [ ] Yes, but I will need assistance and will schedule time during your [office hours](https://calendly.com/fivetran-solutions-team/fivetran-solutions-team-office-hours) for guidance.
- [ ] No.

PR #96, opened 2023-11-06
Deprecation/remove basic transformation tables

fivetran-jamie:
## PR Overview
https://github.com/fivetran/dbt_fivetran_log/issues/93

1.3.0 as this is very breaking for the 2 customers still using basic SQL transformations (and anyone looking at the stale tables for historical analyses) 

**Please detail what change(s) this PR introduces and any additional information that should be known during the review of this PR:**

This PR removes the `transformation` and `trigger_table` sources, staging models, variables, documentation, and downstream transforms (essentially the entire `fivetran_platform__transformation_status` end model)

It does NOT remove `transformation_id` and transformation-oriented logic from the staing [log](https://github.com/fivetran/dbt_fivetran_log/blob/main/models/staging/stg_fivetran_platform__log.sql) model. I chose not to remove this for 2 reasons:
1. the log table is historical, so if someone had basic SQL transformations in the past they will have transformation records in `LOG`. We should continue to correctly label these
2. i believe the product team intends to rollout a new dbt-transformation version of the `transformation` and `trigger_table` tables next year, so `log.transformation_id` could become relevant again (and isn't causing trouble)

### Basic Validation
- [x] dbt compile
- [x] dbt run –full-refresh
- [x] dbt run
- [ ] dbt test -- getting a uniqueness failure in the audit_table on incremental runs... that's unrelated (on main 😓) but looking into it.... 
- [x] dbt run –vars (if applicable)

- [x] The appropriate issue has been linked and tagged
- [x] You are assigned to the corresponding issue and this PR

### Detailed Validation
Please acknowledge that the following validation checks have been performed prior to marking this PR as "ready for review":
- [x] You have validated these changes and assure this PR will address the respective Issue/Feature.
- [x] You are reasonably confident these changes will not impact any other components of this package or any dependent packages.
- [x] You have provided details below around the validation steps performed to gain confidence in these changes.

See the Hex notebook (in height ticket) for my sanity check validations, but truthfully the validation for this task is largely non-sql. Transformations and triggers were only involved in the now-deprecated `transformation_status` end model. No other end models [depend](https://fivetran.github.io/dbt_fivetran_log/#!/model/model.fivetran_log.stg_fivetran_platform__trigger_table?g_v=1&g_i=%2Bstg_fivetran_platform__trigger_table%2B%20%26%26%20%2Bstg_fivetran_platform__transformation%2B) on these sources.

So outside of SQL and into the CLI...

On main, if you do not have `fivetran_platform_using_transformations ` and `fivetran_platform_using_triggers` set to `False` and you don't have the transformation and trigger_table, a `dbt run` would result in
image

One must set these variables to False explicitly (they are true by default) for a successful run
image

image

Thus, this PR makes so it that we achieve the above image without having users (who almost certainly don't have these tables and/or don't want to transform stale data) set any variables.

So if i comment these back out
image

And dbt run, everything works out
image

### Standard Updates
Please acknowledge that your PR contains the following standard updates:
- Package versioning has been appropriately indexed in the following locations:
    - [x] indexed within dbt_project.yml
    - [x] indexed within integration_tests/dbt_project.yml
- [x] CHANGELOG has individual entries for each respective change in this PR
    
- [x] README updates have been applied (if applicable)
    
- [ ] DECISIONLOG updates have been updated (if applicable)
- [x] Appropriate yml documentation has been added (if applicable)

### dbt Docs
Please acknowledge that after the above were all completed the below were applied to your branch:
- [ ] docs were regenerated (unless this PR does not include any code or yml updates) -- waiting to do so on release branch

🥶


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# dbt_fivetran_log v1.3.0
+
+## Breaking Changes
```
fivetran-joemarkiewicz, line 3: Let's add those little alarms here

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+- Deprecated the `transformation` and `trigger_table` source tables and any downstream transforms. These tables only housed information on Fivetran Basic SQL Transformations, which were sunset last year ([PR #96](https://github.com/fivetran/dbt_fivetran_log/pull/96)).
```
fivetran-joemarkiewicz, line 4: We should probably mention the end model in particular that is being removed as well.

Additionally, let's call out the variable being removed for now as well.

fivetran-joemarkiewicz (approved):
LGTM with just a few minor comments to add a bit more to the CHANGELOG

PR #114, opened 2024-02-12
Bug/redshift json parse

fivetran-catfritz:
## PR Overview

### Basic Validation
- [ ] dbt run –full-refresh && dbt test
- [ ] dbt run (if incremental models are present) && dbt test

- [ ] The appropriate issue has been linked, tagged, and properly assigned
- [ ] All necessary documentation and version upgrades have been applied
- [ ] docs were regenerated (unless this PR does not include any code or yml updates)
- [ ] BuildKite integration tests are passing
- [ ] Detailed validation steps have been provided below

### Detailed Validation

:dancer:


macros/fivetran_log_lookback.sql
```diff
@@ -0,0 +1,15 @@
...
+
+{% macro default__fivetran_log_lookback(from_date, datepart='day', interval=7, default_start_date='2010-01-01')  %}
+
+    coalesce(
+        (select {{ dbt.dateadd(datepart=datepart, interval=-interval, from_date_or_timestamp=from_date) }} 
+            from {{ this }}), 
+        {{ "'" ~ default_start_date ~ "'" }}
+        )
+
+{% endmacro %}
```
fivetran-catfritz, line 15: I realize this could have been done in-model, but created the macro in anticipation of adding to fivetran_utils

models/fivetran_platform__audit_table.sql
```diff
@@ -2,10 +2,10 @@
     materialized='incremental',
     unique_key='unique_table_sync_key',
     partition_by={
-        'field': 'sync_start',
-        'data_type': 'timestamp',
-        'granularity': 'day'
+        'field': 'sync_start_day',
+        'data_type': 'date'
```
fivetran-catfritz, line 6: Changed to use the sync_start_day since we have the field and will be easier for use with the lookback macro.

models/fivetran_platform__connector_status.sql
```diff
@@ -1,28 +1,38 @@
...
     from {{ ref('stg_fivetran_platform__log') }}
     where transformation_id is null
+),
 
+parse_json as (
+    select
+        *,
+        {{ fivetran_log.fivetran_log_json_parse(string="filtered_message_data", string_path=["status"]) }} as log_status,
+        {{ fivetran_log.fivetran_log_json_parse(string="filtered_message_data", string_path=["reason"]) }} as log_reason
+    from transformation_removal
```
fivetran-catfritz, line 17: Added this cte to handling all parsing needs in one place. Intention to make this easier to maintain and read.

macros/fivetran_log_lookback.sql
fivetran-joemarkiewicz: I agree and appreciate the forward thinking!

fivetran-joemarkiewicz (changes requested):
@fivetran-catfritz Thanks so much for working through this update!! I have only a few final questions to be addressed below. Coincidentally, when testing these changes I ran into a real world scenario where I had late arriving sync events from Fivetran. If we were using the old method those late events would have been missed; however, with this update they were captured successfully!

Additionally, one last request would be to doubly confirm with the original customers data that the JSON fix proposal will address this issue. I have 99% confidence that this does, but it would be great to fully validate that the compiled code works as expected with the customers data.

Let me know if you have any questions!

models/fivetran_platform__audit_table.sql
fivetran-catfritz: [The default granularity](https://docs.getdbt.com/reference/resource-configs/bigquery-configs#using-table-partitioning-and-clustering) is `day`, and since we're now using a date column instead of timestamp, I thought it would be cleaner to remove the granularity config, but I'm not against adding it back if it would be helpful!

macros/fivetran_log_lookback.sql
```diff
@@ -0,0 +1,15 @@
+{% macro fivetran_log_lookback(from_date, datepart='day', interval=7, default_start_date='2010-01-01') %}
+
+{{ adapter.dispatch('fivetran_log_lookback', 'fivetran_log') (from_date, datepart='day', interval=7, default_start_date='2010-01-01') }}
+
+{%- endmacro %}
+
+{% macro default__fivetran_log_lookback(from_date, datepart='day', interval=7, default_start_date='2010-01-01')  %}
+
+    coalesce(
```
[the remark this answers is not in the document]
fivetran-catfritz: True. I trialed removing this and it seems to compile and run fine. However for now I will leave this pending discussion on Tuesday.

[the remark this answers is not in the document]
fivetran-joemarkiewicz: Thanks, I just wanted to make sure there likely wouldn't be a scenario where this field would actually be `null`. I don't believe including it will hurt so I am comfortable leaving it in.

fivetran-joemarkiewicz (approved):
@fivetran-catfritz thanks again for working through these updates! This PR looks good to go!

fivetran-avinash (commented):
lgtm

Issue #118, opened 2024-03-12
[Bug] String/List Agg Fields in Connector Status model result in error if they exceed limits


fivetran-joemarkiewicz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

It has been reported by a number of users that the following fields in the `fivetran_platform__connector_status` model have the potential to result in a runtime failure if the string/list agg is too long and exceeds the limits. This has been reported on Postgres, Redshift, and Snowflake warehouses:
- `errors_since_last_completed_sync`
- `warnings_since_last_completed_sync`

I had considered the following approaches to address this error:
- ❌ Truncate the string/list agg if it is too long. However, I did not like this approach as while this would result in success, it would be truncating the data without the customer knowing. That would not be an ideal solution.
- 🤔 Fully remove the problematic fields. This is a viable option and we already remove these fields for sql server. However, I would like to explore a better solution without removing these fields entirely.
- 👷 Changing the form and function of these fields to be metrics as opposed to string/list agg. I currently like this solution the best as we don't remove the fields and just change their use to ideally be more useful for all customers and the names will be more consistent. I will explore building this out first. If this is not successful then I will switch to the option of removing the fields entirely.

### Relevant error log or model output

```shell
Error: String '(LISTAGG result)' is too long and would be truncated
```

### Expected behavior

The transformation run does not result in failure

### dbt Project configurations

Quickstart

### Package versions

v1.6.0

### What database are you using dbt with?

postgres, redshift, snowflake

### dbt Version

1.7.1

### Are you willing to open a PR to help address this issue?

- [ ] Yes.
- [ ] Yes, but I will need assistance and will schedule time during our [office hours](https://calendly.com/fivetran-solutions-team/fivetran-solutions-team-office-hours) for guidance
- [ ] No.

fivetran-joemarkiewicz:
This has been addressed in the recent release. As such, closing out this issue.

PR #119, opened 2024-03-14
bugfix/string-agg-removal

fivetran-joemarkiewicz:
## PR Overview
**This PR will address the following Issue/Feature:** Issue #118 

**This PR will result in the following new package version:** `v1.7.0`

This will be a breaking change as we are removing fields from an end model and could cause downstream errors with their removal.

### 🚨 Breaking Changes 🚨: Bug Fixes
- The following fields have been deprecated (removed) as these fields proved to be problematic across warehouses due to the end size of the fields.
  - `errors_since_last_completed_sync`
  - `warnings_since_last_completed_sync`
> Note: If you found these fields to be relevant, you may still reference the error/warning messages from within the underlying `log` table.
- The `fivetran_platform_using_sync_alert_messages` variable has been removed as it is no longer necessary due to the above changes.

### Feature Updates
- The following fields have been added to display the number of error/warning messages sync last completed sync. These fields are intended to substitute the information from deprecated fields listed above.
  - `number_errors_since_last_completed_sync`
  - `number_warnings_since_last_completed_sync`

### Basic Validation

- [X] docs were regenerated (unless this PR does not include any code or yml updates)
- [X] BuildKite integration tests are passing

### Detailed Validation

Please see the corresponding Height ticket for a link to the Hex document that includes the Hex validations notebook.

🔢


models/fivetran_platform__connector_status.sql
```diff
@@ -198,12 +198,9 @@ final as (
...
         connector_recent_logs.last_sync_completed_at,
         connector_recent_logs.set_up_at,
-        coalesce(schema_changes.number_of_schema_changes_last_month, 0) as number_of_schema_changes_last_month
-        
-        {% if var('fivetran_platform_using_sync_alert_messages', true) and target.type != 'sqlserver' %}
-        , {{ fivetran_utils.string_agg("distinct case when connector_recent_logs.event_type = 'SEVERE' then connector_recent_logs.message_data else null end", "'\\n'") }} as errors_since_last_completed_sync
-        , {{ fivetran_utils.string_agg("distinct case when connector_recent_logs.event_type = 'WARNING' then connector_recent_logs.message_data else null end", "'\\n'") }} as warnings_since_last_completed_sync
-        {% endif %}
+        coalesce(schema_changes.number_of_schema_changes_last_month, 0) as number_of_schema_changes_last_month,
+        count(distinct case when connector_recent_logs.event_type = 'SEVERE' then connector_recent_logs.message_data else null end) as number_errors_since_last_completed_sync,
```
fivetran-reneeli, line 202: Will `message_data` always be different across each log record for each connector_id? What happens if there's 2 of the same message_data -- is it intended that they will only be counted once as the same error or warning?

models/fivetran_platform.yml
```diff
@@ -53,17 +53,11 @@ models:
...
+      - name: number_errors_since_last_completed_sync
+        description: Numeric metric indicating the number of error messages raised sync the last sync completion.
 
-      - name: warnings_since_last_completed_sync
-        description: >
-          Aggregated line-separated list of warning messages (in JSON format) raised synce the last 
-          sync completion. Included by default for non-SQL Server targets, but can be disabled by setting
-          the `fivetran_platform_using_sync_alert_messages` var to False.
+      - name: number_warnings_since_last_completed_sync
+        description: Numeric metric indicating the number of warning messages raised sync the last sync completion.
```
fivetran-reneeli, line 60:
```suggestion
        description: Numeric metric indicating the number of warning messages raised since the last sync completion.
```

models/fivetran_platform.yml
```diff
@@ -53,17 +53,11 @@ models:
         description: > 
           The number of schema changes (creating tables or schemas and altering tables) in the past 30 days.
 
-      - name: errors_since_last_completed_sync
-        description: >
-          Aggregated line-separated list of error messages (in JSON format) raised synce the last 
-          sync completion. Included by default for non-SQL Server targets, but can be disabled by setting
-          the `fivetran_platform_using_sync_alert_messages` var to False.
+      - name: number_errors_since_last_completed_sync
+        description: Numeric metric indicating the number of error messages raised sync the last sync completion.
```
fivetran-reneeli, line 57:
```suggestion
      - name: number_errors_since_last_completed_sync
        description: Numeric metric indicating the number of error messages raised since the last sync completion.
```

README.md
```diff
@@ -66,7 +66,7 @@ Include the following Fivetran Platform package version range in your `packages.
 ```yaml
 packages:
   - package: fivetran/fivetran_log
-    version: [">=1.6.0", "<1.7.0"]
+    version: [">=1.7.0", "<1.8.0"]
 ```
 
 > Note that altough the source connector is now "Fivetran Platform", the package retains the old name of "fivetran_log".
```
fivetran-reneeli, line 72:
```suggestion
> Note that although the source connector is now "Fivetran Platform", the package retains the old name of "fivetran_log".
```

fivetran-reneeli (changes requested):
Reviewed with a few suggestions/ question!

models/fivetran_platform__connector_status.sql
fivetran-joemarkiewicz: Good question, in this case we are just using the `message_data` field to count the warning/error entries. The distinct case is not being applied to the message_data, but instead the event_type. So if two message_data of the exact same verbiage are found in the warning event_type, then both will still be counted as a warning since last sync.

fivetran-joemarkiewicz:
Thanks @fivetran-reneeli for the quick and thoughtful review! I just addressed your comments and committed your suggestions. Thanks for catching my spelling mistakes 🤦. I also addressed the open question you had and will hold off regenerating the docs until approval.

fivetran-reneeli:
Thanks @fivetran-joemarkiewicz , just responded above!

models/fivetran_platform__connector_status.sql
fivetran-joemarkiewicz: Ah yes you're right! I was trying to understand why we used distinct previously, but that was to ensure we do not aggregate the same string over and over again. However, like you mentioned we do in fact want to still count those as error warnings with this new field. I will update to remove the distinct clause in the count statement.

fivetran-joemarkiewicz: Thanks for catching this and continuing to dive into this query! Just updated.

fivetran-joemarkiewicz:
@fivetran-reneeli thanks again! I agree with your comment and just applied the changes and regenerated the docs. Let me know if you have any other comments.

models/fivetran_platform__connector_status.sql
fivetran-reneeli: Gotcha, thanks for digging into why we had it in place before! Looks good!

fivetran-reneeli (approved):
lgtm

fivetran-jamie (approved):

PR #122, opened 2024-04-04
testfix/sql-warehouse-schema

fivetran-joemarkiewicz:
This is a patch fix to ensure the integration tests succeed as expected for the SQL Warehouse Datarbricks runtime.

No other changes are necessary as this will be included in the upcoming `v1.7.1` release.


fivetran-reneeli (approved):
Confirming that the sqlw_tests var is set for all commands in that `if` statement!

Issue #120, opened 2024-04-01
[Bug] Databricks SQL Warehouse Compatibility


fivetran-joemarkiewicz:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

If a user is leveraging the a Databricks SQL Warehouse, then they will see the below error. This is due to the incremental strategy `insert_overwrite` not being compatibile with Databricks SQL Warehouse runtimes.

### Relevant error log or model output

```shell
spark.sql.sources.partitionOverwriteMode is not available
```

### Expected behavior

The incremental strategy works regardless of Databricks runtime.

### dbt Project configurations

Any

### Package versions

Latest

### What database are you using dbt with?

databricks

### dbt Version

Any

### Additional Context

We likely will consider the following since we cannot differentiate between databricks runtimes in dbt when using adapter calls.

- Leverage the `merge` strategy with some failsafes in place to avoid the downsides of the merge strategy

- Remove the incremental strategy

### Are you willing to open a PR to help address this issue?

- [x] Yes.
- [ ] Yes, but I will need assistance and will schedule time during our [office hours](https://calendly.com/fivetran-solutions-team/fivetran-solutions-team-office-hours) for guidance
- [ ] No.

fivetran-joemarkiewicz:
Providing an update here - we have been able to find a way to differentiate between Databricks SQL Warehouse and Databricks All Purpose Clusters in the above PR! I had gone back and forth on either leveraging the `merge` strategy or just opting to remove the incremental strategy for SQL Warehouse runtimes. As we are not 100% confident in the `merge` incremental strategy for the `fivetran_platform__audit_table` model at this time, we are moving forward with making this model **not** incremental for SQL Warehouse Databricks runtimes.

PR #123, opened 2024-04-09
bugfix/databricks-sql-path-adjustment

fivetran-joemarkiewicz:
## PR Overview
**This PR will address the following Issue/Feature:** Issue #120 

**This PR will result in the following new package version:** `v1.7.2`

This will be non breaking as it will only adjust the regex search in the Datarbricks SQL Warehouse to remove a leading backslash. There will be no breaking changes resulting from this release.

> ## Bug Fixes
> - Removal of the leading `/` from the `target.http_path` regex search within the `is_databricks_sql_warehouse()` macro to accurately identify SQL Warehouse Databricks destinations in Quickstart.
>   - The macro above initially worked as expected in dbt core environments; however, in Quickstart implementations this data model was not working. This was due to Quickstart removing the leading `/` from the `target.http_path`. Thus resulting in the regex search to always fail. 

### Basic Validation

### Detailed Validation

In order to test this I did the following:
- Understood why the `v1.7.1` release is not working as expected. It is due to code on the Quickstart side which removes the leading `/` from the `target.http_path`. Please see internal messages for link to code. Therefore, removing the leading `/` from the regex search in the `is_databricks_sql_warehouse()` macro should do the trick with working on both dbt core **and** Quickstart.
  - I received confirmation from engineering that this should work as expected.
- After applying the update to the macro I tested to ensure the logic still works as expected on a Databricks SQL Warehouse destination. You can see in the below screenshot that after running `dbt run -t databricks-sql` (after a full refresh to simulate if there was any incremental strategy) there was no incremental logic included in the model. Further, no incremental models were captured in the end complete statement.
  - image
- To ensure the logic is still working as expected for the Databricks All Purpose Cluster, I performed the same steps to ensure the incremental strategy was being applied. I was able to confirm this is the case!
  -  image
- Lastly, I wanted to ensure that there is no change for a non Databricks destination. While no changes were made that should impact any other destination, I did still want to check at least two. I checked Snowflake and BigQuery. Both looked to be in order and utilizing the incremental strategy on subsequent loads!
 - BigQuery
   - image
 - Snowflake
   - image
 

◀️


fivetran-reneeli (approved):
Got it, so there was a leading `/` in the SQL Warehouse regex causing it to never match to Databricks because that in the backend, they removed that leading `/` in http_path. And the other warehouses were not impacted 👍  Looks good!

fivetran-avinash (approved):
@fivetran-joemarkiewicz good to merge!

Issue #124, opened 2024-05-06
[Bug] `fivetran_log_json_parse.string_path` arg should be an array in `fivetran_platform__audit_user_activity` model


fivetran-jamie:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

On [line 5](https://github.com/fivetran/dbt_fivetran_log/blob/d3556141f93b7635b8ed55fea10c0dc0e590994d/models/fivetran_platform__audit_user_activity.sql#L5C9-L5C110) we have 
`{{ fivetran_log.fivetran_log_json_parse(string='message_data', string_path='actor') }} as actor_email`

But it should be 
`{{ fivetran_log.fivetran_log_json_parse(string='message_data', string_path=['actor']) }} as actor_email`

This is how the macro is used [everywhere else](https://github.com/search?q=repo%3Afivetran%2Fdbt_fivetran_log+fivetran_log_json_parse+language%3ASQL+path%3A%2F%5Emodels%5C%2F%2F&type=code).

This causes the json parsing to not output anything and therefore limit the output of the `fivetran_platform__audit_user_activity` model

### Relevant error log or model output

```shell
Compiled fivetran_platform__audit_user_activity code for Snowflake:
try_parse_json(message_data)['a']['c']['t']['o']['r'] as actor_email

For Postgres:
  case when message_data ~ '^\s*[\{].*[\}]?\s*$' -- Postgres has no native json check, so this will check the string for indicators of a JSON object
    then message_data::json #>> '{a,c,t,o,r}'
    else null end

For BQ:
json_extract_scalar(message_data, '$.a.c.t.o.r')

etc.
```

### Expected behavior

I would expect the macro to compile to:

**Snowflake**
`try_parse_json(message_data)['actor'] as actor_email`

**Postgres**
 `case when message_data ~ '^\s*[\{].*[\}]?\s*$' -- Postgres has no native json check, so this will check the string for indicators of a JSON object
    then message_data::json #>> '{actor}'
    else null end` 

**BQ**
`json_extract_scalar(message_data, '$.actor')`

### dbt Project configurations

na

### Package versions

na

### What database are you using dbt with?

snowflake

### dbt Version

na

### Additional Context

https://fivetran.slack.com/archives/C02919TN9AT/p1714746313138259

### Are you willing to open a PR to help address this issue?

- [X] Yes.
[...]

PR #126, opened 2024-05-11
Bug/postgres performance

fivetran-catfritz:
## PR Overview
- Internal ticket
- #124 

- v1.7.3 non-breaking since there are no schema changes and results are the same as prior

> ## Performance Improvements
> - Updated the sequence of JSON parsing for model `fivetran_platform__audit_table` to reduce runtime. 
> 
> ## Bug Fixes
> - Updated model `fivetran_platform__audit_user_activity` to correct the JSON parsing used to determine column `email`.  
> 
> ## Under the hood
> - Updated logic for macro `fivetran_log_lookback` to align with logic used in similar macros in other packages. 

### Basic Validation
- [x] dbt run –full-refresh && dbt test
- [x] dbt run (if incremental models are present) && dbt test

- [x] The appropriate issue has been linked, tagged, and properly assigned
- [x] All necessary documentation and version upgrades have been applied

### Detailed Validation

- See internal ticket for data quality validation

- Runtime comparison:

  - Note: this was tested on a subset of the `log` data limited to about 250k rows. This was done by filtering the staging log model  using filter

```sql
where cast(created_at as date) < cast('2024-05-10' as date)
and cast(created_at as date) >= cast('2024-05-07' as date)
```

which produces these results:

v1.7.2 run: ~ 1hr
Screenshot 2024-05-11 at 2 17 09 AM

v1.7.3 run: ~ 10 mins
Screenshot 2024-05-11 at 1 00 29 AM

🌳


macros/fivetran_log_json_parse.sql
```diff
@@ -28,7 +28,7 @@
 {% macro postgres__fivetran_log_json_parse(string, string_path) %}
 
   case when {{ string }} ~ '^\s*[\{].*[\}]?\s*$' -- Postgres has no native json check, so this will check the string for indicators of a JSON object
-    then {{ string }}::json #>> '{ {%- for s in string_path -%}{{ s }}{%- if not loop.last -%},{%- endif -%}{%- endfor -%} }'
+    then {{ string }}::jsonb #>> '{ {%- for s in string_path -%}{{ s }}{%- if not loop.last -%},{%- endif -%}{%- endfor -%} }'
```
fivetran-joemarkiewicz, line 31: I know this is incredibly small, but we should call this out as a change in the CHANGELOG. Also, I am fairly certain this won't be breaking, but can you confirm that this datatype change won't cause any datatype changes for Postgres users? I tested myself and didn't see any issues when running prod then dev right after, but just want to double check.

fivetran-joemarkiewicz (approved):
@fivetran-catfritz Thanks for investigating and working through this update! It is also great to see the performance improvements it brought with our sample data set. I am hopeful this will save runtimes for our customers using Postgres. 

I do have one small callout to include a change in the CHANGELOG and for your eyes on my thoughts around a changing datatype in the json parse macro. Those are relatively small and I am comfortable approving this release. Let me know if you have any questions around my comment. Thanks!

macros/fivetran_log_json_parse.sql
fivetran-catfritz: Added to under the hood. I also tested this and it would make sense there is no issue since the end results in either version are strings, and the cast to json is an intermediate step.

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+# dbt_fivetran_log v1.7.3
+[PR #126](https://github.com/fivetran/dbt_fivetran_log/pull/126) includes the following updates:
+
+## Performance Improvements
+- Updated the sequence of JSON parsing for model `fivetran_platform__audit_table` to reduce runtime. 
+
+## Bug Fixes
+- Updated model `fivetran_platform__audit_user_activity` to correct the JSON parsing used to determine column `email`.
```
fivetran-jamie, line 8: could be good to note that was causing `fivetran_platform__audit_user_activity` to potentially have 0 rows

fivetran-jamie (approved):

Issue #128, opened 2024-05-23
[Feature] Update docs to make clarify sync times 


fivetran-reneeli:
### Is there an existing feature request for this?

- [X] I have searched the existing issues

### Describe the Feature

The sync times in FIVETRAN_PLATFORM__AUDIT_TABLE are not the most recent sync times that are found in the source table because we only take record when a table is being written to. See [this CTE](https://github.com/fivetran/dbt_fivetran_log/blob/5fef3647a47e682236713c274895905ec5a81643/models/fivetran_platform__audit_table.sql#L89-L93) in that model that has the event_subtype limiter. 

In conclusion, if there is no corresponding event logged that a table is written to, the sync time doesn't get persisted (that’s why the most recent sync times are present in the staging log table but not in our audit model since we only show the timestamps for when write events occur).

We should make this more clear for the fields in the docs for FIVETRAN_PLATFORM__AUDIT_TABLE related to sync times.

### Are you interested in contributing this feature?

- [X] Yes.
- [ ] Yes, but I will need assistance and will schedule time during your [office hours](https://calendly.com/fivetran-solutions-team/fivetran-solutions-team-office-hours) for guidance.
- [ ] No.

fivetran-joemarkiewicz:
This has been addressed in the latest v1.8.0 release!

PR #116, opened 2024-03-06
feature/singular-tests-addition

fivetran-joemarkiewicz:
## PR Overview
**This PR will address the following Issue/Feature:** No corresponding issue or feature

**This PR will result in the following new package version:** `v1.6.0`

**Please provide the finalized CHANGELOG entry which details the relevant changes included in this PR:**

### Bug Fixes
- It was observed that the `fivetran_platform__audit_table` model could potentially introduce duplicates for database like connectors. To address this issue, the join logic in the end model was adjusted to no longer be determined by date ranges and instead be defined via model generated indexes. This ensures true uniqueness in the end model.

### Testing Updates
- Addition of the `tests/` folder which includes various custom validation tests to ensure the the output of the end models are matching our expectations.
  - These tests are disabled by default as they are not necessary to run after every execution of the models. If you would like to enabled these tests you may activate them by defining the `fivetran_validation_tests_enabled` variable as `true`.

### Basic Validation
Please acknowledge that you have successfully performed the following commands locally:
- [ ] dbt run –full-refresh && dbt test
- [ ] dbt run (if incremental models are present) && dbt test

Before marking this PR as "ready for review" the following have been applied:
- [ ] The appropriate issue has been linked, tagged, and properly assigned
- [ ] All necessary documentation and version upgrades have been applied
- [ ] docs were regenerated (unless this PR does not include any code or yml updates)
- [ ] BuildKite integration tests are passing
- [ ] Detailed validation steps have been provided below

### Detailed Validation
Please share any and all of your validation steps:

### If you had to summarize this PR in an emoji, which would it be?
 
🖥️

PR #130, opened 2024-05-29
feature/databricks-delta-incremental-support

fivetran-joemarkiewicz:
## PR Overview
**This PR will address the following Issue/Feature:** Internal tickets and Issue #128 

**This PR will result in the following new package version:** `v1.8.0`

When I tested this locally for Databricks there was actually no error when running without a full refresh. However, the table format did not change. Therefore, a breaking change should be leveraged to ensure a full refresh is ran and the delta table format is applied.

> ## 🚨 Breaking Changes 🚨
> > ⚠️ Since the following changes result in the table format changing, we recommend running a `--full-refresh` after upgrading to this version to avoid possible incremental failures.
> - For Databricks All Purpose clusters the `fivetran_platform__audit_table` model will now be materialized using the delta table format (previously parquet). 
>   - Delta tables are generally more performant than parquet and are also more widely available for Databricks users. Previously, the parquet file format was causing compilation issues on customers managed tables.
> 
> ## Documentation Updates
> - Updated the `sync_start` and `sync_end` field descriptions for the `fivetran_platform__audit_table` to explicitly define that these fields only represent the sync start/end times for when the connector wrote new or modified existing records to the specified table.
> 
> ## Under the Hood
> - The `is_databricks_sql_warehouse` macro has been renamed to `is_databricks_all_purpose` and has been modified to return `true` if the Databricks runtime being used is an all purpose cluster (previously this macro checked if a sql warehouse runtime was used).
>   - This update was applied as there have been other Databricks runtimes discovered (ie. an endpoint and external runtime) which do not support the `insert-overwrite` incremental strategy used in the `fivetran_platform__audit_table` model. 
> - In addition to the above, for Databricks users the `fivetran_platform__audit_table` model will now leverage the incremental strategy only if the Databricks runtime is all purpose. Otherwise, all other Databricks runtimes will not leverage an incremental strategy.

### Basic Validation

### Detailed Validation

To validate these changes the validation tests were included and you can see they were successful for the following destinations:
- BigQuery
image

- Databricks All Purpose Cluster
image

- Databricks SQL Warehouse
image

Additionally, I validated that the All Purpose Cluster appropriately runs an incremental strategy and the non All Purpose (SQL Warehouse in this case) does not run an incremental strategy.
- Databricks All Purpose Cluster
image

- Databricks SQL Warehouse
image

Finally, I confirmed that the Delta format runs as expected and without issue on the Databricks All Purpose cluster on incremental runs.
image

🌳


fivetran-joemarkiewicz:
The SQL Server buildkite test is currently failing, but that is due to a permission issue which should hopefully be resolved soon. I will re-kick off the integration tests once that is resolved. I don't imagine SQL Server would be failing for any of these changes; therefore, this should be good to review even with the failing buildkite test.

integration_tests/tests/consistency/consistency__audit_table.sql
```diff
@@ -0,0 +1,67 @@
...
+
+-- Checking to ensure the dev totals match the prod totals
+consistency_check as (
+    select *
+    from final_consistency_check
+    where prod_total != dev_total
+),
+
+-- The current release changes the row count of the audit table model intentionally.
+-- The below queries prove the records that do not match are still accurate by checking the source.
```
fivetran-catfritz, line 45: @fivetran-joemarkiewicz Quick question--are these tests in anticipation of the model changes from [this PR](https://github.com/fivetran/dbt_fivetran_log/pull/116) being merged?

fivetran-catfritz: @fivetran-joemarkiewicz Got it thanks! Maybe changing the comment to something like "For use when the current release..." would solve any confusion.

fivetran-joemarkiewicz: I like that suggestion! Just made the update.

fivetran-catfritz (approved):
Changes lgtm, and I was able to give it a full refresh and incremental run in Databricks. All looks good there, so approved!

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+# dbt_fivetran_log v1.8.0
+[PR #130](https://github.com/fivetran/dbt_fivetran_log/pull/130) includes the following updates:
+
+## 🚨 Breaking Changes 🚨
+> ⚠️ Since the following changes result in the table format changing, we recommend running a `--full-refresh` after upgrading to this version to avoid possible incremental failures.
+- For Databricks All Purpose clusters the `fivetran_platform__audit_table` model will now be materialized using the delta table format (previously parquet). 
+  - Delta tables are generally more performant than parquet and are also more widely available for Databricks users. Previously, the parquet file format was causing compilation issues on customers managed tables.
```
fivetran-avinash, line 7:
```suggestion
  - Delta tables are generally more performant than parquet and are also more widely available for Databricks users. Previously, the parquet file format was causing compilation issues on customers' managed tables.
```

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+# dbt_fivetran_log v1.8.0
+[PR #130](https://github.com/fivetran/dbt_fivetran_log/pull/130) includes the following updates:
+
+## 🚨 Breaking Changes 🚨
+> ⚠️ Since the following changes result in the table format changing, we recommend running a `--full-refresh` after upgrading to this version to avoid possible incremental failures.
+- For Databricks All Purpose clusters the `fivetran_platform__audit_table` model will now be materialized using the delta table format (previously parquet).
```
fivetran-avinash, line 6:
```suggestion
- For Databricks All-Purpose clusters, the `fivetran_platform__audit_table` model will now be materialized using the delta table format (previously parquet). 
```

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+## 🚨 Breaking Changes 🚨
+> ⚠️ Since the following changes result in the table format changing, we recommend running a `--full-refresh` after upgrading to this version to avoid possible incremental failures.
+- For Databricks All Purpose clusters the `fivetran_platform__audit_table` model will now be materialized using the delta table format (previously parquet). 
+  - Delta tables are generally more performant than parquet and are also more widely available for Databricks users. Previously, the parquet file format was causing compilation issues on customers managed tables.
+
+## Documentation Updates
+- Updated the `sync_start` and `sync_end` field descriptions for the `fivetran_platform__audit_table` to explicitly define that these fields only represent the sync start/end times for when the connector wrote new or modified existing records to the specified table.
+
+## Under the Hood
+- The `is_databricks_sql_warehouse` macro has been renamed to `is_databricks_all_purpose` and has been modified to return `true` if the Databricks runtime being used is an all purpose cluster (previously this macro checked if a sql warehouse runtime was used).
```
fivetran-avinash, line 13:
```suggestion
- The `is_databricks_sql_warehouse` macro has been renamed to `is_databricks_all_purpose` and has been modified to return `true` if the Databricks runtime being used is an all-purpose cluster (previously this macro checked if a sql warehouse runtime was used).
```

README.md
fivetran-avinash, line 1: Unrelated to the PR, was checking to ensure the Databricks README instructions were up to date, and I think it looks good, but noticed there are two Databricks Dispatch Configurations on lines 48-54 and lines 115-121. Do we need both?

models/fivetran_platform.yml
```diff
@@ -234,9 +234,9 @@ models:
       - name: write_to_table_end
         description: Timestamp of when Fivetran finished writing to this table.
       - name: sync_start
-        description: Timestamp of when this connector-level sync began. 
+        description: Timestamp of when this connector-level sync began and wrote or modified records to the table.
       - name: sync_end
-        description: Timestamp of when this connector-level sync finished, if successful. 
+        description: Timestamp of when this connector-level sync finished, if successful and wrote or modified records to the table.
```
fivetran-avinash, line 239:
```suggestion
        description: Timestamp of when this connector-level sync finished, if successful, and wrote or modified records to the table. 
```

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+
+## 🚨 Breaking Changes 🚨
+> ⚠️ Since the following changes result in the table format changing, we recommend running a `--full-refresh` after upgrading to this version to avoid possible incremental failures.
+- For Databricks All Purpose clusters the `fivetran_platform__audit_table` model will now be materialized using the delta table format (previously parquet). 
+  - Delta tables are generally more performant than parquet and are also more widely available for Databricks users. Previously, the parquet file format was causing compilation issues on customers managed tables.
+
+## Documentation Updates
+- Updated the `sync_start` and `sync_end` field descriptions for the `fivetran_platform__audit_table` to explicitly define that these fields only represent the sync start/end times for when the connector wrote new or modified existing records to the specified table.
+
+## Under the Hood
```
fivetran-avinash, line 12: Are the consistency and integrity tests under the hood? Or should we not call them out?

fivetran-joemarkiewicz: Yeah they are under the hood - would be good to include in the CHANGELOG. Will update!

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+- The `is_databricks_sql_warehouse` macro has been renamed to `is_databricks_all_purpose` and has been modified to return `true` if the Databricks runtime being used is an all purpose cluster (previously this macro checked if a sql warehouse runtime was used).
+  - This update was applied as there have been other Databricks runtimes discovered (ie. an endpoint and external runtime) which do not support the `insert-overwrite` incremental strategy used in the `fivetran_platform__audit_table` model.
```
fivetran-avinash, line 14:
```suggestion
  - This update was applied as there have been other Databricks runtimes discovered (ie. an endpoint and external runtime) which do not support the `insert_overwrite` incremental strategy used in the `fivetran_platform__audit_table` model. 
```

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+
+## Documentation Updates
+- Updated the `sync_start` and `sync_end` field descriptions for the `fivetran_platform__audit_table` to explicitly define that these fields only represent the sync start/end times for when the connector wrote new or modified existing records to the specified table.
+
+## Under the Hood
+- The `is_databricks_sql_warehouse` macro has been renamed to `is_databricks_all_purpose` and has been modified to return `true` if the Databricks runtime being used is an all purpose cluster (previously this macro checked if a sql warehouse runtime was used).
+  - This update was applied as there have been other Databricks runtimes discovered (ie. an endpoint and external runtime) which do not support the `insert-overwrite` incremental strategy used in the `fivetran_platform__audit_table` model. 
+- In addition to the above, for Databricks users the `fivetran_platform__audit_table` model will now leverage the incremental strategy only if the Databricks runtime is all purpose. Otherwise, all other Databricks runtimes will not leverage an incremental strategy.
```
fivetran-avinash, line 15:
```suggestion
- In addition to the above, for Databricks users the `fivetran_platform__audit_table` model will now leverage the incremental strategy only if the Databricks runtime is all-purpose. Otherwise, all other Databricks runtimes will not leverage an incremental strategy.
```

README.md
fivetran-joemarkiewicz: Great catch! The lines on 115-121 are in fact just a duplicate of the lines on 48-54. I removed the instructions at the bottom of the README and updated the CHANGELOG to reflect this. Thanks for noticing!

CHANGELOG.md
fivetran-joemarkiewicz: Added

models/fivetran_platform__audit_table.sql
```diff
@@ -1,13 +1,13 @@
 {{ config(
-    materialized='table' if is_databricks_sql_warehouse(target) else 'incremental',
+    materialized='incremental' if is_databricks_all_purpose(target) else 'table',
```
fivetran-avinash, line 2: Just double checking this logic, since the conditions have been flipped:

if `is_databricks_all_purpose(target)` is true, then it'll be materialized as incremental.
If it's false, then it'll be a table. Which makes sense for databricks.

However, what would this entail for the other warehouses? I'm looking at the macro loop and it seems that it would only be true for the case when it's a databricks runtime is all-purpose. But that would be false for the other warehouses, so they would now be materialized as tables instead of incremental. Is that the intention?

fivetran-avinash (commented):
@fivetran-joemarkiewicz Looks like you handled all the minor tweaks already, so just one question on the new audit table config before approving.

models/fivetran_platform__audit_table.sql
fivetran-joemarkiewicz: @fivetran-avinash BRILLIANT catch! You are exactly correct and this was not the intention. This will now set the databricks all-purpose cluster to use the incremental strategy and also turn it off for non all-purpose clusters.... BUT ALSO turns the incremental strategy off for all other warehouses 😱. 

Extremely thankfully you reviewed this and caught this gap. Let me revisit the code and account for all other warehouses.

fivetran-joemarkiewicz: I just made some code updates to account for the above issue. @fivetran-avinash @fivetran-catfritz would you be able to review and let me know if you have any questions or there are any other considerations to take into account?

See below for validations that the materializations are working as expected on each platform.
- ✅  BigQuery (subsequent runs use incremental strategy)
image

- ✅  Snowflake (subsequent runs use incremental strategy)
image

- ✅  Redshift (subsequent runs use incremental strategy)
image

- ✅  Postgres (subsequent runs use incremental strategy)
image

- ✅  Databricks All-Purpose Cluster (subsequent runs use incremental strategy)
image

- ✅  Databricks SQL Warehouse (no runs use the incremental strategy)
image

CHANGELOG.md
fivetran-avinash: This was added under Documentation Updates. Make sure to move it down!

CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+  - Delta tables are generally more performant than parquet and are also more widely available for Databricks users. Previously, the parquet file format was causing compilation issues on customers' managed tables.
+
+## Documentation Updates
+- Updated the `sync_start` and `sync_end` field descriptions for the `fivetran_platform__audit_table` to explicitly define that these fields only represent the sync start/end times for when the connector wrote new or modified existing records to the specified table.
+- Addition of integrity and consistency validation tests within integration tests for every end model.
+- Removed duplicate Databricks dispatch instructions listed in the README.
+
+## Under the Hood
+- The `is_databricks_sql_warehouse` macro has been renamed to `is_incremental_compatible` and has been modified to return `true` if the Databricks runtime being used is an all-purpose cluster (previously this macro checked if a sql warehouse runtime was used) **or** if any other non Databricks supported destination is being used.
```
fivetran-avinash, line 15:
```suggestion
- The `is_databricks_sql_warehouse` macro has been renamed to `is_incremental_compatible` and has been modified to return `true` if the Databricks runtime being used is an all-purpose cluster (previously this macro checked if a sql warehouse runtime was used) **or** if any other non-Databricks supported destination is being used.
```

fivetran-joemarkiewicz: Applied in latest commit

fivetran-avinash (approved):
@fivetran-joemarkiewicz New updates look good!

Only small call-out is if we feel there are any warehouses we set up that might not support incremental materialization in the future, we might want to explicitly do an `elif target.type 
 in ('bigquery', 'snowflake', etc.)` --> true, else false, just for full coverage. But that is not a present concern and can be revisited if we add more destinations. 

A few additional recommended edits in the Changelog but otherwise lgtm.

fivetran-catfritz:
@fivetran-avinash really really good catch. I ran this only in Databricks so definitely didn't catch that then! 

@fivetran-joemarkiewicz Tagging on to Avinash's comments, a more future-proof way to handle the logic update might be:
1. Put the macro back to just check if it's runtime.
2. Set the config like:
```sql
...
materialized='table' if target.type == 'databricks' 
  and not is_databricks_runtime(or whatever the old version name was)() else 'incremental'
...
```
That way we don't have to list out the other warehouses. What do you think?

fivetran-joemarkiewicz:
@fivetran-catfritz I like that idea, but the benefit of listing out each of the warehouses is we are explicitly only using the incremental strategy if we know the destination is supported. If it is not in our supported list, then we use the `table` materialization. This likely will provide the greatest opportunity for success if for some reason a destination not supported is used to run this model.

fivetran-catfritz:
@fivetran-joemarkiewicz Makes sense--in that case approved on my end!

PR #154, opened 2025-06-13
feature/update-freshness-config

fivetran-catfritz:
## PR Overview 
**Package version introduced in this PR:** 
 - v2.2.0
 
**This PR addresses the following Issue/Feature(s):**

- #151 

**Summary of changes:**  

- Updates freshness to align with the new config framework introduced in dbt Core v1.9.5. Documents how users on older versions of dbt Core will/can interface with freshness

### Submission Checklist  
- [ ] Alignment meeting with the reviewer (if needed)  
  - [ ] Timeline and validation requirements discussed  
- [x] Provide validation details:  
  - [x] **Validation Steps:** Check for unintentional effects (e.g., add/run consistency & integrity tests)
  - [x] **Testing Instructions:** Confirm the change addresses the issue(s)
  - [ ] **Focus Areas:** Complex logic or queries that need extra attention  

- When < dbt v1.9.5, no tests are run. 
  - Screenshot 2025-06-13 at 12 13 20 PM

- When >= dbt v1.9.5, test run successfully. 
  - Screenshot 2025-06-13 at 12 38 14 PM

### Changelog  

- [x] Draft changelog for PR  
- [ ] Final changelog for release review

PR #156, opened 2025-06-18
Update deprecated config for freshness

b-per:
Move freshness under config to make it compatible with Fusion and remove deprecations in dbt-core 1.10


fivetran-joemarkiewicz:
Thanks for opening this @b-per and highlighting the necessary changes for dbt-core 1.10! We've applied these same updates within PR #154. This is planning to go live Monday in a new release.

PR #193, opened 2026-05-21
feature/new_fivetran_log_end_model

fivetran-savage:
- [GA-1013639](https://fivetran.atlassian.net/browse/GA-1013639)
- Adds `fivetran_platform__transformation_run_log` to the package to provide data on number of models run per transformation, whether the run succeeded/failed/partially succeeded, run duration, as well as the associated destination, connection(s), and full run log.


models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,188 @@
+-- depends_on: {{ ref('stg_fivetran_platform__log') }}
+-- depends_on: {{ ref('stg_fivetran_platform__transformation_runs') }}
+-- depends_on: {{ ref('stg_fivetran_platform__destination') }}
+-- depends_on: {{ ref('stg_fivetran_platform__connection') }}
```
fivetran-jamie: Are these necessary to include?

.quickstart/quickstart.yml
```diff
@@ -24,5 +24,6 @@ public_models: [
   "fivetran_platform__connection_status",
   "fivetran_platform__usage_history",
   "fivetran_platform__audit_table",
-  "fivetran_platform__mar_table_history"
+  "fivetran_platform__mar_table_history",
+  "fivetran_platform__transformation_run_log"
```
fivetran-jamie, line 32: Should the `fivetran_platform_using_transformations` variable be incorporated into the `table_variables` in this file?

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,188 @@
...
+{% if var('fivetran_platform_using_transformations', does_table_exist('transformation_runs')) %}
```
fivetran-jamie: I know this is the format we currently use in the staging model, but I think our standard approach for disabling/enabling models might be better here. So instead of the if/else null + does_table_exist setup: 
```sql
{{ config(enabled=var('fivetran_platform_using_transformations', true) }} 
```
And `fivetran_platform_using_transformations` would be dynamically set in Quickstart

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,188 @@
...
+
+),
+
+transformation_logs as (
+
+    select
+        event_subtype,
+        message_data as full_run_log,
+        {{ fivetran_log.fivetran_log_json_parse(string="message_data", string_path=["id"]) }} as transformation_id,
+        cast({{ fivetran_log.fivetran_log_json_parse(string="message_data", string_path=["startTime"]) }} as {{ dbt.type_timestamp() }}) as started_at,
```
fivetran-jamie, line 37: Seeing some null `started_at` values when running this on our internal data. Looks like sometimes it's nested within `results.stepResults` -- we might wanna coalesce with this? 

Also in these cases, I see that there is an additional `endTime` within `results.stepResults` and is slightlyyyy different (0 - a few seconds earlier) from the not-nested `endTime` value. Should we prioritize using this value in a coalesce?

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,188 @@
...
+        cast({{ fivetran_log.fivetran_log_json_parse(string="message_data", string_path=["result", "stepResults", 0, "failedModelRuns"]) }} as {{ dbt.type_int() }}) as failed_model_runs
+    from logs
+    where event_subtype like 'transformation%'
+        and event_subtype != 'transformation_start'
+    --transformation failure and success events capture start/end datetimes so the start event is redundant
+),
+
+connection_ids_unnested as (
+
+    {{ fivetran_log.fivetran_log_connection_ids_unnested() }}
```
fivetran-jamie, line 52: Looks like sometimes `integrations` is empty, so the connection info is missing for some quickstart runs. However, we could probably still figure out which connections they belong to based on the job name, which I think is always `<Connector type>/<connection/schema name>`

What do you think about parsing that as a backup? We could get connection name easily, and perhaps use that + destination_id to join in the `connection` table to get connection_id

models/fivetran_platform.yml
```diff
@@ -193,6 +193,47 @@ models:
...
+      - name: connection_ids
+        description: Comma-separated list of connection IDs scheduled to trigger this transformation.
+      - name: connection_names
+        description: Comma-separated list of connection names scheduled to trigger this transformation.
+      - name: transformation_name
+        description: User-defined name of the transformation.
+      - name: transformation_id
+        description: System-generated unique ID of the transformation job, parsed from the log message.
+      - name: transformation_type
+        description: The type of transformation (e.g., DBT).
```
fivetran-jamie:
```suggestion
        description: The type of transformation (e.g., DBT_CORE, QUICKSTART).
```

These are the values I see in BQ

models/fivetran_platform.yml
```diff
@@ -193,6 +193,47 @@ models:
...
+      - name: successful_model_runs
+        description: Number of models that ran successfully in this transformation execution.
+      - name: failed_model_runs
+        description: Number of models that failed in this transformation execution.
+      - name: transformation_status
+        description: >
+          Outcome of the transformation run. Possible values are `transformation_succeeded`,
+          `transformation_partially_succeeded`, and `transformation_failed`.
+      - name: event_subtype
```
fivetran-jamie: I don't think this field is included in the model

models/fivetran_platform__transformation_run_log.sql
fivetran-jamie: Actually I don't anticipate this being an issue -- I'm only seeing this on an old (though still active) internal fivetran log connection. Will leave my previous comment in just to call this out a little

fivetran-jamie (commented):
Looking great -- see comments

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: It wasn't running in my local environment and then I added them and it ran, but I'll try again without. Could have been a caching issue or something else. I'll try with a clean schema.

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: Noted.

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: Any idea how big of an issue this is in customer data? The fallback is a little wonky because of the unnest array in the json. The join from job_name will only give us one connection_id, whereas the array has the potential to give us multiple connection_ids. That's probably not the case for most quickstarts but it is worth noting in a decisionlog. Based on querying the two datasets (test connector and ours), it seems that if there are multiple connectors, integrations is always populated, so it seems fine that the fallback only grabs one connection_id, at least based on the data we have.

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: Not sure what the call is here...do we think it's only an issue in the older connection data? I'm thinking I'll put it in the decisionlog so we have a record of it and then if customers flag that they have null values for start_time/duration, we can add the coalesce with stepresults at that point?

macros/fivetran_log_connection_ids_unnested.sql
```diff
@@ -0,0 +1,88 @@
...
+        (transformation_logs.full_run_log::jsonb) -> 'schedule' -> 'integrations'
+    ) as unnested_integrations(connection_id_raw)
+
+{% endmacro %}
+
+{% macro snowflake__fivetran_log_connection_ids_unnested() %}
+
+    select
+        transformation_logs.transformation_id,
+        flattened_integration.value::varchar as connection_id
```
fivetran-jamie, line 23: Is casting to varchar necessary here? I see that we cast to a string downstream (all warehouses) in the `distinct_connection_ids` CTE [here](https://github.com/fivetran/dbt_fivetran_log/blob/feature/new_fivetran_log_end_model/models/fivetran_platform__transformation_run_log.sql#L75)

macros/fivetran_log_connection_ids_unnested.sql
```diff
@@ -0,0 +1,88 @@
...
+        json_extract_array_element_text(
+            json_extract_path_text(transformation_logs.full_run_log, 'schedule', 'integrations'),
+            numbers.n,
+            true
+        ) as connection_id
+    from transformation_logs
+    cross join (
+        select 0 as n union all select 1 union all select 2 union all select 3
+        union all select 4 union all select 5 union all select 6 union all select 7
+        union all select 8 union all select 9
```
fivetran-jamie: Does this mean that if a Redshift user is running one job on 11 different connections, the 11th will be cut out? Or the whole thing will be if its length > 10? 

If so, we should document this somewhere (and if possible confirm that 10 is a safe number to use in terms of # of customers affected)

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,174 @@
...
+        successful_model_runs,
+        failed_model_runs,
+        case
+            when step_success = 'true' then 'transformation_succeeded'
+            when step_success = 'false'
+                and successful_model_runs > 0
+                and failed_model_runs > 0
+                then 'transformation_partially_succeeded'
+            else 'transformation_failed'
+        end as transformation_status,
```
fivetran-jamie, line 200: If `step_success` turns out null, we label it as `transformation_failed`. I wonder if something like `unknown` would be most appropriate in these cases (if they exist)

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,174 @@
...
+        full_run_log
+    from with_duration
+
+)
+
+select *
```
fivetran-jamie, line 206: What do you think about adding a surrogate key? Could be called `run_id` and hashed on `transformation_id` and `started_at` and/or `ended_at`

We would be able to apply a `unique` test to this in the yml doc like the audit_table's `unique_table_sync_key`

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,174 @@
...
+        job_destinations.job_id,
+        job_destinations.destination_id,
+        job_destinations.destination_name,
+        cast(connections.connection_id as {{ dbt.type_string() }}) as fallback_connection_id,
+        cast(connections.connection_name as {{ dbt.type_string() }}) as fallback_connection_name
+    from job_destinations
+    left join connections
+        on job_destinations.project_type = 'QUICKSTART'
+        and job_destinations.job_name like '%/%'
+        and connections.connection_name = {{ dbt.split_part(string_text='job_destinations.job_name', delimiter_text="'/'", part_number=2) }}
```
fivetran-jamie: For performance, it could be good to move some of this to an earlier CTE rather than the join conditions. The first two can be put in a where clause in `job_destinations` (only used here right?), and we can also do the `split_part` there

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,174 @@
...
+    select *
+    from {{ ref('stg_fivetran_platform__log') }}
+
+),
+
+transformation_runs as (
+
+    select *
+    from {{ ref('stg_fivetran_platform__transformation_runs') }}
+
```
fivetran-jamie, line 14: It looks like we're currently only using this table for the connection_id/name fallback. Could be good to include this more generally (especially since it enables this model). The `project_type`, `model_runs`, and maybe `free_type` columns seem appropriate to include in this model

models/fivetran_platform__transformation_run_log.sql
fivetran-jamie: I like your fallback -- we're providing data as complete as we can

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: At this point, we're getting the transformation status entirely from `step_success`, and since our original goal was to re-create the Quickstart UI but with the full history, the only statuses are success, failure, and partial success. I can add another criteria for failed_model_runs > 0 and successful_model_runs = 0 for a failure status, and then add an else 'unknown'?

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: Probably a good idea for this model, will add.

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: I'll add project_type and free_type for sure. I've got `successful_model_runs` and `failed_model_runs` from the parsed json already...do you think we need to bring in another `model_runs` column?

fivetran-savage: If the idea with that is to track model runs for usage/cost purposes, the `fivetran_platform__usage_history` model does that already.

macros/fivetran_log_connection_ids_unnested.sql
fivetran-savage: It's necessary because of the way lateral_flatten works in Snowflake. Snowflake's lateral flatten returns its value column as a VARIANT, which shows up in a select with double quotes, so you have to cast as varchar to extract a clean scalar string without the quotes. The downstream CTE cast as string is just a safety net to normalize everything and could maybe be removed, but I'm always nervous with nested JSONS because there are a lot of ways the data can get wonky.

macros/fivetran_log_connection_ids_unnested.sql
fivetran-savage: I means that if it has 11 connections, the 11th would be cut out and it would only show 10. Looks like Claude picked 10 arbitrarily in my first pass, and after a deeper dive, we've got multiple accounts with over 30 connections, so I've changed it to accommodate up to 100. Let me know if you think this is overkill.

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: Made it its own CTE since the fallback only applies to Quickstart jobs.

models/fivetran_platform__transformation_run_log.sql
fivetran-jamie: Yeah I think that'd be a good fallback just in case! Do we need to check the `failed_model_runs` and `successful_model_runs` for failure, or can we just do:

```sql
        case
            when step_success = 'true' then 'transformation_succeeded'
            when step_success = 'false'
                and successful_model_runs > 0
                and failed_model_runs > 0
                then 'transformation_partially_succeeded'
            when step_success = 'false' then 'transformation_failed'
            else 'unknown'
        end as transformation_status,
```

models/fivetran_platform__transformation_run_log.sql
fivetran-jamie: Ah good point -- yeah let's add just project_type and free_type then!

macros/fivetran_log_connection_ids_unnested.sql
fivetran-jamie: I think that's good! We can lower that in the future if Redshift performance is suffering

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: Per our conversation this morning, holding off on surrogate keys until we have a larger conversation with the team.

fivetran-jamie (approved):

PR #194, opened 2026-05-21
release_branch/v2.6.0

fivetran-savage:
## PR Overview 
**Package version introduced in this PR:** 
- 2.6.0

**This PR addresses the following Issue/Feature(s):**

- [GA-1020307](https://fivetran.atlassian.net/browse/GA-1020307)

**Summary of changes:**  

- Adds `stg_fivetran_platform__connector_sdk_log` to package.
- Adds net-new end model, `fivetran_platform__errors_and_warnings`, that allows users to track severity levels of errors and warnings for both sdk and standard connections.
- Adds `fivetran_platform__transformation_run_log` to the package to provide data on number of models run per transformation, whether the run succeeded/failed/partially succeeded, run duration, as well as the associated destination, connection(s), and full run log.
- Adds `fivetran_platform__sync_metrics model`, which returns one record per completed ssync_stats log event, combining each sync's extract, process, load timing and volume statistics with its total duration and total records modified, enriched with connection and destination details
- Adds `stg_fivetran_platform__audit_trail` to the package, which records user actions performed within your Fivetran account (what was changed, by whom, and how). Only available for customers on the Enterprise plan and above.
- Adds `fivetran_platform__audit_trail_enriched`, which enriches each audit trail event with the names of the primary and secondary resources involved (resolved for `CONNECTION`, `DESTINATION`, `ACCOUNT`, and `USER` resources) and the details of the user who performed the action. Only available for customers on the Enterprise plan and above.
- Adds the optional `fivetran_platform_log_start_date` variable to limit the amount of data included in log-based models. When set to a date (`YYYY-MM-DD`), `stg_fivetran_platform__log` filters to records on or after that date.
- Changes materializes `stg_fivetran_platform__log` as table.

### Submission Checklist  
- [x] Alignment meeting with the reviewer (if needed)  
  - [x] Timeline and validation requirements discussed  
- [ ] Provide validation details:  
  - [x] **Validation Steps:** Check for unintentional effects (e.g., add/run consistency & integrity tests)
  - [x] **Testing Instructions:** Confirm the change addresses the issue(s)
  - [x] **Focus Areas:** Complex logic or queries that need extra attention  
- [x] Merge any relevant open PRs into this PR

### Changelog  

- [x] Draft changelog for PR  
- [x] Final changelog for release review


integration_tests/ci/test_scenarios.yml
fivetran-joemarkiewicz, line 1: Let's also add tests to check the following variables not using the default state:
- fivetran_platform_using_transformations
- fivetran_platform_using_audit_trail

models/staging/stg_fivetran_platform__audit_trail.sql
fivetran-joemarkiewicz, line 1: Why are we using the `does_table_exist()` macro here? Wouldn't the config enabled ([example](https://github.com/fivetran/dbt_fivetran_log/blob/0da7ac69669322de4b4c9ea88b5a0071b60b23e7/models/staging/stg_fivetran_platform__user.sql#L1)) serve better here to just turn off the model if the table isn't used as opposed to creating an empty table with null recrods.

models/staging/src_fivetran_platform.yml
```diff
@@ -245,6 +271,41 @@ sources:
           - name: _fivetran_synced
             description: Timestamp when the record was last synced.
 
+      - name: audit_trail
+        identifier: "{{ var('fivetran_platform_audit_trail_identifier', 'audit_trail') }}"
+        description: >
+          Table of user actions performed within your Fivetran account, including what was changed and how.
+          This table is only available on the Enterprise plan and above.
+        config:
```
fivetran-joemarkiewicz, line 279: If the variable is set to false we will want to disable this source.

dbt_project.yml
```diff
@@ -10,6 +10,10 @@ models:
     staging:
       +schema: stg_fivetran_platform
       +materialized: view
+      stg_fivetran_platform__log:
+        +materialized: table
+      tmp:
+        +materialized: view
```
fivetran-joemarkiewicz: This is not the behavior we ultimately want. Instead we want the default materialization to be `stg_fivetran_platform__log_tmp` = view and `stg_fivetran_platform__log` will still remains a view as well.

Then, if the `fivetran_platform_log_start_date` is set then we will change the materialization of `stg_fivetran_platform__log_tmp` to be table. The `stg_fivetran_platform__log` will remain a view.

We'll want this approach because:
1. We don't want to change the existing behavior. Not all users want to materialize the tmp/staging model as a table as this is an added storage cost. With this build, if a user doesn't set the variable then they are essentially storing all the log data in duplicative locations. We want the default to be a view.
2. If the purpose of this update is to improve performance, then we want the very first model (the tmp model) to materialize as a table. This will then ensure all downstream models are only using the data from that table. If we materialize the tmp model as a view it would result in the staging model still scanning the entire log source table each time.

DECISIONLOG.md
```diff
@@ -25,3 +25,32 @@ vars:
...
+
+## Redshift connection limit in `fivetran_platform__transformation_run_log`
+The `connection_ids` and `connection_names` fields are derived by unnesting the `schedule.integrations` array from the run log. Most warehouses use a native lateral-unnest function with no cap, but Redshift has no such primitive, so the `fivetran_log_connection_ids_unnested` macro iterates the array indices with a generated numbers table that produces `0-99` — capping connections per transformation job at **100 on Redshift only**. A job referencing more than 100 connections has the overflow silently dropped (the run still appears, just with an incomplete connection list). We chose 100 based on production data: as of 06/2026 the largest Redshift job referenced 32 connections and only 19 jobs (9 destinations) exceeded 10, so 100 leaves comfortable headroom. If the limit is ever reached, the generator can be extended (for example, by adding a hundreds digit to the cross join).
```
fivetran-joemarkiewicz:
```suggestion
The `connection_ids` and `connection_names` fields are derived by unnesting the `schedule.integrations` array from the run log. Most warehouses use a native lateral-unnest function with no cap, but Redshift has no such primitive, so the `fivetran_log_connection_ids_unnested` macro iterates the array indices with a generated numbers table that produces `0-99` — capping connections per transformation job at **100 on Redshift only**. A job referencing more than 100 connections has the overflow silently dropped (the run still appears, just with an incomplete connection list). We chose 100 as this should leave comfortable headroom.
```

DECISIONLOG.md
```diff
@@ -25,3 +25,32 @@ vars:
...
+
+## Null `started_at` for older transformation runs
+The `started_at` and `ended_at` fields in `fivetran_platform__transformation_run_log` are parsed from the top-level `startTime` and `endTime` keys in the run log. Older transformation run events do not include these top-level keys; instead, the timestamps are nested under `result.stepResults[0]`. For those older runs, `started_at` (and therefore `duration`) may be null, even though the run did capture a start time.
+
+We have chosen not to address this for now, since it only affects older/stale data. If customers report null `started_at` or `duration` values, we can coalesce the top-level keys with the nested `result.stepResults[0].startTime` and `result.stepResults[0].endTime` values to recover them.
```
fivetran-joemarkiewicz:
```suggestion
The `started_at` and `ended_at` fields in `fivetran_platform__transformation_run_log` are parsed from the top-level `startTime` and `endTime` keys in the run log. Older transformation run events do not include these top-level keys; instead, the timestamps are nested under `result.stepResults[0]`. For those older runs, `started_at` (and therefore `duration`) may be null, even though the run did capture a start time. We have chosen not to address this for now, since it only affects older/stale data.
```

DECISIONLOG.md
```diff
@@ -25,3 +25,32 @@ vars:
...
+
+## Connector SDK log source
+Not all customers have the `connector_sdk_log` source table, as it is only populated for accounts using the Fivetran Connector SDK. This source and its downstream logic are controlled by the `fivetran_platform_using_connector_sdk_log` variable, which defaults to `true`. If your destination does not contain the `connector_sdk_log` table, set this variable to `false` in your root `dbt_project.yml` to disable the source and exclude Connector SDK events from `fivetran_platform__errors_and_warnings`:
+
+```yml
+vars:
+  fivetran_platform_using_connector_sdk_log: false ## Default is true. Set to false if the connector_sdk_log source table is not present in your destination.
+```
```
fivetran-joemarkiewicz: This belongs in the README, not in the DECISIONLOG

README.md
```diff
@@ -142,6 +154,18 @@ vars:
     fivetran_platform_<default_table_name>_identifier: your_table_name 
 ```
 
+#### Limit the Lookback Window
+By default, log-based models scan all available log history from the source `log` table. On large tables, this can increase query costs and/or run times. To limit the scan window, set the `fivetran_platform_log_start_date` variable to the earliest date you want to include (in `YYYY-MM-DD` format):
+```yml
+vars:
+    fivetran_platform_log_start_date: '2024-01-01' # scan only log history on or after this date
```
fivetran-joemarkiewicz:
```suggestion
    fivetran_platform_log_start_date: '2024-01-01' # scan only log history on and after this date
```

README.md
```diff
@@ -142,6 +154,18 @@ vars:
...
+```
+This filter is applied in `stg_fivetran_platform__log`. Because it is a fixed date rather than a rolling window, the boundary does not move over time — a `--full-refresh` always reloads from the same start date, so historical data is never lost.
+
+Incremental downstream models accumulate history over time on regular runs, and their `--full-refresh` output is bounded by the start date you set. As these models grow, you can periodically move `fivetran_platform_log_start_date` forward to keep the full-refresh scan (and the staging table) manageable. Run `dbt run --full-refresh` after setting or changing this variable to ensure all models reflect the updated window.
+
+This affects the following models: `fivetran_platform__connection_status`, `fivetran_platform__schema_changelog`, `fivetran_platform__audit_user_activity`, `fivetran_platform__connection_daily_events`, and `fivetran_platform__audit_table`.
```
fivetran-joemarkiewicz: This doesn't render very well. Small adjustment to format this in markdown
```suggestion
This affects the following models: 
- `fivetran_platform__connection_status`
- `fivetran_platform__schema_changelog`
- `fivetran_platform__audit_user_activity`
- `fivetran_platform__connection_daily_events`
- `fivetran_platform__audit_table`.
```

CHANGELOG.md
```diff
@@ -1,3 +1,31 @@
...
+| [stg_fivetran_platform__audit_trail](https://fivetran.github.io/dbt_fivetran_log/#!/model/model.fivetran_log.stg_fivetran_platform__audit_trail) | New staging model |  |  | Records user actions performed within your Fivetran account (what was changed, by whom, and how). Only available for customers on the Enterprise plan and above. |
+| [stg_fivetran_platform__connector_sdk_log](https://fivetran.github.io/dbt_fivetran_log/#!/model/model.fivetran_log.stg_fivetran_platform__connector_sdk_log) | New staging model |  |  | Staging model for the new `connector_sdk_log` source table. |
+| [stg_fivetran_platform__log](https://fivetran.github.io/dbt_fivetran_log/#!/model/model.fivetran_log.stg_fivetran_platform__log) | Materialization | view | table | Now materialized as a table to apply the `fivetran_platform_log_start_date` filter after column transformations, reducing source scan costs on downstream models for large log tables. |
```
fivetran-joemarkiewicz: This needs to be revised based off my earlier comment

CHANGELOG.md
```diff
@@ -1,3 +1,31 @@
...
+
+## Bug Fix
+- Fixes SQL Server JSON array-index paths in the `fivetran_log_json_parse` macro.
+
+## Feature Update
+- Adds the `fivetran_platform_using_connector_sdk_log` variable (default `true`) to enable or disable the `connector_sdk_log` source and its downstream logic. Set this to `false` if your destination does not contain the `connector_sdk_log` table.
```
fivetran-joemarkiewicz: Remove this and instead mention the variable in the schema/data change table above where we mention the the staging model.

Also, the other variable for audit_trail needs to be included as well.

CHANGELOG.md
```diff
@@ -1,3 +1,31 @@
...
+
+- Adds the optional `fivetran_platform_log_start_date` variable to control how much source log data is scanned/loaded. When set to a date (`YYYY-MM-DD`), `stg_fivetran_platform__log` filters to log records on or after that date. As downstream incremental models grow, you can periodically move the start date forward to keep full-refresh scans manageable. See the [README](https://github.com/fivetran/dbt_fivetran_log/blob/main/README.md#limit-the-lookback-window) for details. Affects `fivetran_platform__connection_status`, `fivetran_platform__schema_changelog`, `fivetran_platform__audit_user_activity`, `fivetran_platform__connection_daily_events`, and `fivetran_platform__audit_table` (**note:** run `dbt run --full-refresh` after setting or changing this variable).
```
fivetran-joemarkiewicz: Clarify early in this entry what the default state of this variable is.

CHANGELOG.md
```diff
@@ -1,3 +1,31 @@
+# dbt_fivetran_log v2.6.0
+
+[PR #201](https://github.com/fivetran/dbt_fivetran_log/pull/201) includes the following updates:
```
fivetran-joemarkiewicz:
```suggestion
[PR #194](https://github.com/fivetran/dbt_fivetran_log/pull/194) includes the following updates:
```

models/staging/stg_fivetran_platform__audit_trail.sql
```diff
@@ -0,0 +1,78 @@
...
+            )
+        }}
+    from base
+),
+
+field_conversion as (
+    select
+        *,
+        {{ fivetran_log.json_to_string("old_values", source_columns_in_relation) }} as old_values_string,
+        {{ fivetran_log.json_to_string("new_values", source_columns_in_relation) }} as new_values_string
```
fivetran-joemarkiewicz, line 26: If "new_values" and "old_values" are de-selected at the connector level this model will fail. We'll need to apply this macro only after we know the fields exist which we can guarantee after the fill_staging_columns macro results.

models/fivetran_platform__audit_trail_enriched.sql
```diff
@@ -0,0 +1,137 @@
...
+        select
+            connection_id,
+            connection_name,
+            row_number() over (
+                partition by connection_id
+                order by is_deleted asc, set_up_at desc
+            ) as nth_connection_record
+        from {{ ref('stg_fivetran_platform__connection') }}
+    ) deduped
+    where nth_connection_record = 1
```
fivetran-joemarkiewicz: To make this easier to maintain going forward, let's break this into two ctes instead of a subquery

models/fivetran_platform__audit_trail_enriched.sql
```diff
@@ -0,0 +1,137 @@
...
+{% endif %}
+
+final as (
+
+    select
+        audit_trail.unique_audit_trail_key,
+        audit_trail.audit_trail_id,
+        audit_trail.captured_at,
+        audit_trail.user_id,
+{%- if var('fivetran_platform_using_user', true) %}
```
fivetran-joemarkiewicz: Format this jinja so it's on the same line as the fields.

models/fivetran_platform__audit_trail_enriched.sql
```diff
@@ -0,0 +1,137 @@
...
+    from audit_trail
+
+{%- if var('fivetran_platform_using_user', true) %}
+
+    -- actor who performed the action
+    left join users as actor
+        on audit_trail.user_id = actor.user_id
+{% endif %}
+
+    -- resolve the primary resource (the container) to its name
```
fivetran-joemarkiewicz: What's the container?

models/fivetran_platform__errors_and_warnings.sql
```diff
@@ -0,0 +1,82 @@
...
+        'connector_sdk' as connector_type
+    from {{ ref('stg_fivetran_platform__connector_sdk_log') }}
+    where lower(level) in ('warning', 'error', 'severe')
+        and connection_id is not null
+),
+{%- endif %}
+
+unioned as (
+
+    select * from log_events
```
fivetran-joemarkiewicz:
```suggestion
    select * 
    from log_events
```

models/fivetran_platform__errors_and_warnings.sql
```diff
@@ -0,0 +1,82 @@
...
+
+    {% if var('fivetran_platform_using_connector_sdk_log', true) -%}
+    union all
+    select * from connector_sdk_events
```
fivetran-joemarkiewicz:
```suggestion
    union all
    
    select * 
    from connector_sdk_events
```

models/fivetran_platform__connection_errors_and_warnings.sql
```diff
@@ -0,0 +1,82 @@
...
+        connection_id,
+        connection_name
+    from (
+        select
+            connection_id,
+            connection_name,
+            row_number() over (partition by connection_id order by connection_name) as connection_row
+        from {{ ref('stg_fivetran_platform__connection') }}
+    ) deduped
+    where connection_row = 1
```
fivetran-joemarkiewicz, line 68: Same comment as earlier note.

models/fivetran_platform__connection_errors_and_warnings.sql
fivetran-joemarkiewicz, line 1: If this is only capturing connection errors and warnings should we specify that in the name of the model?

models/fivetran_platform__sync_metrics.sql
```diff
@@ -0,0 +1,107 @@
...
+    from (
+        select
+            *,
+            row_number() over (
+                partition by connection_id
+                order by is_deleted asc, set_up_at desc
+            ) as nth_connection_record
+        from {{ ref('stg_fivetran_platform__connection') }}
+    ) deduped
+    where nth_connection_record = 1
```
fivetran-joemarkiewicz, line 68: Same note to split this into two ctes for easier maintainability

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,230 @@
...
+            connection_id,
+            connection_name,
+            destination_id,
+            row_number() over (
+                partition by connection_id
+                order by is_deleted asc, set_up_at desc
+            ) as nth_connection_record
+        from {{ ref('stg_fivetran_platform__connection') }}
+    ) deduped
+    where nth_connection_record = 1
```
fivetran-joemarkiewicz, line 45: Same note to break this into two ctes for easier maintainability

models/staging/stg_fivetran_platform__connector_sdk_log.sql
```diff
@@ -0,0 +1,34 @@
+{{ config(enabled=var('fivetran_platform_using_connector_sdk_log', true)) }}
```
fivetran-joemarkiewicz: We should probably default this variable to `false`. For existing core users if they don't have this enabled then they would encounter a failure after upgrading. Let's let users enable this, but have the default be false.

fivetran-joemarkiewicz: Same note for the other new source tables being added

models/fivetran_platform__connection_errors_and_warnings.sql
fivetran-joemarkiewicz: Also is there a reason we order by connection name?

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,230 @@
...
+        successful_model_runs,
+        failed_model_runs,
+        case
+            when step_success = 'true' then 'transformation_succeeded'
+            when step_success = 'false'
+                and successful_model_runs > 0
+                and failed_model_runs > 0
+                then 'transformation_partially_succeeded'
+            when step_success = 'false' then 'transformation_failed'
+            else 'unknown'
```
fivetran-joemarkiewicz: Should we filter out these unknown statuses? Or is there a possible miss  somewhere. Looking at some of these full_run_log records with an "unknown" status looks like they are in fact a transformation, but is missing key details.

models/fivetran_platform.yml
```diff
@@ -193,6 +193,63 @@ models:
...
+        description: Total run duration formatted as HH:MM:SS.
+      - name: destination_id
+        description: Foreign key referencing the destination associated with the transformation.
+      - name: destination_name
+        description: Name of the destination associated with the transformation.
+      - name: project_type
+        description: Type of transformation project (e.g., `DBT_CORE`, `QUICKSTART`).
+      - name: free_type
+        description: Indicates the type of free model run, or `PAID` for paid model runs.
+      - name: connection_ids
```
fivetran-joemarkiewicz, line 229: We should also specify here that dbt transformations not associated with a connection (cron scheduled) will have an expected null connection_id

models/fivetran_platform__transformation_run_log.sql
fivetran-joemarkiewicz: Are these ultimately deleted jobs?

models/fivetran_platform__errors_and_warnings.sql
```diff
@@ -0,0 +1,82 @@
+with log_events as (
+
+    select
+        log_id as event_id,
+        connection_id,
+        sync_id,
+        created_at as event_time,
+        event_type as severity_level,
+        message_data as message,
+        'standard_connector' as connector_type
```
fivetran-joemarkiewicz: The name `connector_type` could be misleading. When I see this I may think "google_sheets" as this is how other existing models use the `connector_type` field name.

Let's consider renaming this to something more specific to this use case. Maybe a boolean like `is_custom_connector` or something like that? Open to other ideas, but we should change the name to not cause confusion with other uses of this field name.

models/fivetran_platform__transformation_run_log.sql
fivetran-joemarkiewicz: If so, we should document in the yml

models/fivetran_platform__connection_errors_and_warnings.sql
```diff
@@ -0,0 +1,82 @@
...
+        unioned.connection_id,
+        connection.connection_name,
+        unioned.event_time,
+        unioned.severity_level,
+        unioned.message,
+        unioned.connector_type,
+        unioned.sync_id
+    from unioned
+    left join connection
+        on unioned.connection_id = connection.connection_id
```
fivetran-joemarkiewicz, line 90: This left join means we'll bring in records for deleted connections. We should do this, but let's document that deleted connections will exist, but will have null connection_name

models/fivetran_platform__sync_metrics.sql
```diff
@@ -0,0 +1,107 @@
...
+        sync_stats.date_day,
+        sync_stats.day_of_week,
+        sync_stats.extract_time_s,
+        sync_stats.extract_volume_mb,
+        sync_stats.process_time_s,
+        sync_stats.process_volume_mb,
+        sync_stats.load_time_s,
+        sync_stats.load_volume_mb,
+        sync_stats.total_time_s,
+        records_modified.row_count
```
fivetran-joemarkiewicz: I noticed this field looks to be null more than I would think. Is that expected?

models/fivetran_platform__audit_trail_enriched.sql
```diff
@@ -0,0 +1,137 @@
...
+        audit_trail.primary_resource_type,
+        audit_trail.primary_resource_id,
+        coalesce(
+            primary_connection.connection_name,
+            primary_destination.destination_name,
+            primary_account.account_name
+            {%- if var('fivetran_platform_using_user', true) %},
+            primary_user.email
+            {%- endif %}
+        ) as primary_resource_name,
```
fivetran-joemarkiewicz, line 82: I noticed the TEAM resource type always has a null value for these primary and secondary values. This is expected since we don't bring in TEAM yet. 

Let's document this, but create a FR for us to bring in TEAM in the future.

fivetran-joemarkiewicz (changes requested):
@fivetran-savage great work on this PR. A few comments before approving.

models/staging/stg_fivetran_platform__audit_trail.sql
fivetran-savage: It was modeled after `stg_fivetran_platform__transformation_runs` which also uses `does_table_exist()`. Should we switch that one over to config enabled as well?

models/fivetran_platform__errors_and_warnings.sql
fivetran-savage: Good point. Let me give it a bit of thought, but I like the `is_custom_connector boolean` idea.

models/fivetran_platform__audit_trail_enriched.sql
fivetran-savage: Missed this one. Claude is always trying to sneak in nested subqueries.

models/staging/stg_fivetran_platform__audit_trail.sql
fivetran-savage: This is applied after the fill_staging_columns macro, in the fields cte that starts on line 11.

models/fivetran_platform__audit_trail_enriched.sql
fivetran-savage: It's a way to explain the hierarchy. 'Container' is the main resource the action is being performed on and 'member' (commented a bit later) is the secondary resource, but it's confusing, so I'll change the comment to be clearer.

models/fivetran_platform__connection_errors_and_warnings.sql
fivetran-savage: Fair point. Will update.

models/fivetran_platform__transformation_run_log.sql
fivetran-savage: This is a label for when the step_success is null, suggested as a fallback in the original review. I was marking these as failed, given that in the UI of the run log there are only failed/succeeded/partially succeed. We could filter them out or document. I'm honestly not sure if the reason they're missing details is that they were ultimately deleted or there is some other reason, hence the 'unknown' status. We do leave the full run log in as a field for the user to parse if they want the full details.

models/fivetran_platform__connection_errors_and_warnings.sql
fivetran-savage: Do you think we should document inline in the model code or put a note in the connection_name description in the yml?

CHANGELOG.md
fivetran-savage: Is "optional" not clear enough?

models/fivetran_platform__sync_metrics.sql
fivetran-savage: By "this field" do you mean row_count? If yes, then that is expected. This model is a combination of "records_modified" and "sync_stats" event subtypes with the left join on sync_stats so as not to lose data from any syncs, and there are lots of syncs where no records were modified so the modified row_count is null. That does beg the question if maybe we should alias that field to "rows_modified" or "modified_row_count" because technically it's not the total row count, it's the count of rows that were modified in some way.

models/fivetran_platform__connection_errors_and_warnings.sql
fivetran-savage: That was an early iteration of the connection dedupe. Will update to match the others. It should be ordered by set_up_at.

README.md
```diff
@@ -142,6 +159,23 @@ vars:
...
+This filter is applied in `stg_fivetran_platform__log`. Because it is a fixed date rather than a rolling window, the boundary does not move over time — a `--full-refresh` always reloads from the same start date, so historical data is never lost.
+
+Incremental downstream models accumulate history over time on regular runs, and their `--full-refresh` output is bounded by the start date you set. As these models grow, you can periodically move `fivetran_platform_log_start_date` forward to keep the full-refresh scan (and the staging table) manageable. Run `dbt run --full-refresh` after setting or changing this variable to ensure all models reflect the updated window.
+
+This affects the following models: 
+  `fivetran_platform__connection_status` 
+  `fivetran_platform__schema_changelog` 
+  `fivetran_platform__audit_user_activity` 
+  `fivetran_platform__connection_daily_events`
+  `fivetran_platform__audit_table`.
```
fivetran-joemarkiewicz:
```suggestion
  - `fivetran_platform__connection_status` 
  - `fivetran_platform__schema_changelog` 
  - `fivetran_platform__audit_user_activity` 
  - `fivetran_platform__connection_daily_events`
  - `fivetran_platform__audit_table`
```

README.md
```diff
@@ -111,6 +115,19 @@ vars:
...
+vars:
+    fivetran_platform_using_audit_trail: true # Default is false. Set to true only if you have the audit_trail source table
+```
+## Connector SDK log source
+Not all customers have the `connector_sdk_log` source table, as it is only populated for accounts using the Fivetran Connector SDK. This source and its downstream logic are controlled by the `fivetran_platform_using_connector_sdk_log` variable, which defaults to `false`. Quickstart automatically enables the source when the `connector_sdk_log` table is present in your schema. For core dbt users, if your destination contains the `connector_sdk_log` table, set this variable to `true` in your root `dbt_project.yml` to enable the source and include Connector SDK events in `fivetran_platform__connection_errors_and_warnings`:
+
+```yml
+vars:
+  fivetran_platform_using_connector_sdk_log: true ## Default is false. Set to true only if the connector_sdk_log source table is present in your destination.
+```
```
fivetran-joemarkiewicz: Consolidate both of these new sections into the existing section above. They all generally say the same thing and can be consolidated.

.quickstart/quickstart.yml
```diff
@@ -24,5 +30,15 @@ public_models: [
...
+  "fivetran_platform__mar_table_history",
+  "fivetran_platform__transformation_run_log",
+  "fivetran_platform__connection_errors_and_warnings",
+  "fivetran_platform__sync_metrics"
 ]
+
+supported_vars:
+  fivetran_platform_log_start_date:
+    type: string
+    description: "Earliest date (YYYY-MM-DD) of log history to include when building log-based models. When unset, all log history is used. Because it is a fixed date, a full refresh always reloads from this start date and never loses history. Affects `fivetran_platform__connection_status`, `fivetran_platform__schema_changelog`, `fivetran_platform__audit_user_activity`, `fivetran_platform__connection_daily_events`, and `fivetran_platform__audit_table`."
```
fivetran-joemarkiewicz: The newly added models in this PR are also impacted by the variable. Since this list is becoming exhaustive, it may be best to instead list which models are not impacted. It looks like it's just two: `fivetran_platform__mar_table_history`, `fivetran_platform__usage_history`

integration_tests/ci/test_scenarios.yml
```diff
@@ -42,4 +42,15 @@ test_scenarios:
...
+    include_incremental: false
+
+  - name: "audit trail enabled"
+    vars:
+      fivetran_platform_using_audit_trail: true
+    include_incremental: false
+
+  - name: "connector sdk log enabled"
+    vars:
+      fivetran_platform_using_connector_sdk_log: true
```
fivetran-joemarkiewicz, line 55: In one of these let's also set the `fivetran_platform_log_start_date` variable to a relevant date in the seed data so we can continually test that functionality going forward as well.

models/fivetran_platform__transformation_run_log.sql
```diff
@@ -0,0 +1,234 @@
+{{ config(enabled=var('fivetran_platform_using_transformations', true))}}
```
fivetran-joemarkiewicz: It's probably safest to set this to `false` by default in order to not cause complications for existing users who may have this variable being set to false. We should also mention we're planning to remove the dynamic feature in the future (currently being used in the staging model).

We also don't document this variable in the README. We need to add that to the enable/disable variables section.

DECISIONLOG.md
fivetran-joemarkiewicz, line 1: Let's add a decisionlog for the variable materialization of the log staging model based on setting the start date variable.

models/fivetran_platform__connection_errors_and_warnings.sql
```diff
@@ -0,0 +1,93 @@
...
+        connection_name
+    from connection_ranked
+    where connection_row = 1
+),
+
+final as (
+
+    select
+        {{ dbt_utils.generate_surrogate_key([
+            'unioned.event_id',
```
fivetran-joemarkiewicz, line 75: I noticed we use event_id in the surrogate key, but we don't actually select it in the final statement. Is this intentional?

fivetran-joemarkiewicz (changes requested):

integration_tests/ci/test_scenarios.yml
fivetran-savage: These are both tables that many customers won't have enabled. Does that matter? I'm inclined to add another test on a table that more customers will have.

models/fivetran_platform__connection_errors_and_warnings.sql
fivetran-savage: It was needed for uniqueness but I don't think it adds value to the output. I can add it back in if you disagree.

README.md
```diff
@@ -103,14 +107,28 @@ vars:
...
+- `connector_sdk_log` is only populated for accounts using the Fivetran Connector SDK. Enabling it includes Connector SDK events in `fivetran_platform__connection_errors_and_warnings`.
+- `transformation_runs` is only populated for accounts using Fivetran Transformations. This variable resolves differently in each layer when you leave it unset:
+    - `stg_fivetran_platform__transformation_runs` has no fixed default. It checks your schema for the `transformation_runs` table at runtime. If the table exists, the staging model and the `paid_model_runs`, `free_model_runs`, and `total_model_runs` fields in `fivetran_platform__usage_history` are populated. If it does not, the staging model builds empty and those fields are null.
+    - `fivetran_platform__transformation_run_log` defaults to `false` and does not build. The runtime check does not apply here, so set this variable to `true` to build the model.
+
+  Setting the variable explicitly overrides both layers. `true` requires the `transformation_runs` table to exist, and `false` disables the staging model and its downstream fields even when the table does exist.
```
fivetran-joemarkiewicz: Was this meant to be a bulleted item? If so, let's add that. If not, let's fix the indent

README.md
```diff
@@ -142,6 +160,23 @@ vars:
...
+This filter is applied in `stg_fivetran_platform__log`. Because it is a fixed date rather than a rolling window, the boundary does not move over time — a `--full-refresh` always reloads from the same start date, so historical data is never lost.
+
+Incremental downstream models accumulate history over time on regular runs, and their `--full-refresh` output is bounded by the start date you set. As these models grow, you can periodically move `fivetran_platform_log_start_date` forward to keep the full-refresh scan (and the staging table) manageable. Run `dbt run --full-refresh` after setting or changing this variable to ensure all models reflect the updated window.
+
+This affects the following models: 
+  `fivetran_platform__connection_status` 
+  `fivetran_platform__schema_changelog` 
+  `fivetran_platform__audit_user_activity` 
+  `fivetran_platform__connection_daily_events`
+  `fivetran_platform__audit_table`
```
fivetran-joemarkiewicz: Can we have this be a bulleted list so it renders better in markdown. Also, I believe we're missing the new models that are also impacted.

fivetran-joemarkiewicz (approved):

README.md
fivetran-savage: It's a continuation of the paragraph started with "transformation_runs has no fixed default" but it looks odd when rendered, so I'll move all the bullets below the paragraph.

PR #197, opened 2026-06-24
feature/add_audit_trail_events

fivetran-savage:
To be merged into release branch 2.6.0
- [GA-1027402](https://fivetran.atlassian.net/browse/GA-1027402)
- Adds `stg_fivetran_platform__audit_trail` to the package, which records user actions performed within your Fivetran account (what was changed, by whom, and how). Only available for customers on the Enterprise plan and above.


models/staging/stg_fivetran_platform.yml
```diff
@@ -184,6 +184,43 @@ models:
...
+  - name: stg_fivetran_platform__audit_trail
+    description: >
+      Table of user actions performed within your Fivetran account, including what was changed and how.
+      This table is only available on the Enterprise plan and above.
+    columns:
+      - name: unique_audit_trail_key
+        description: Surrogate key hashed on `audit_trail_id` and `captured_at` to uniquely identify each audit trail event.
+        tests:
+          - unique
+          - not_null
```
fivetran-catfritz, line 196: I think the not_null test will fail on Redshift with the all nulls + limit 1 scenario, but I don't think we should change anything. Just a note for later in case it does come up.

fivetran-catfritz (approved):
looks great! one small comment but nothing to revise

models/staging/stg_fivetran_platform.yml
fivetran-savage: Fair point. This whole thing will get a big review before release, so we can always remove the not_null test at that point.

